PR #18, opened 2023-01-07
Lighter colors and other UI changes

iamlogand:
# Lighter colors and other UI changes

This change mainly makes the authentication UI lighter in color and more accessible. However, I do like the original darker colors and would like to see them brought back in a "dark theme" somewhere down the road.

## Change List
- Changed the font from Cabin to Open Sans. Cabin was more interesting, but Open Sans is more readable.
- Fixed an issue where only the default font weight was being imported
- Changed the default text color to tyrian purple
- Auth UI changes:
  - Made the auth UI lighter
  - Added borders
  - Added a transition effect to the feedback
  - Added `focus` styling to fields to improve accessibility
  - Fixed a defect where errored fields were not highlighting after submission
  - Increased the font weight of the auth title
  - Increased sizes of elements and gaps between elements
- Changed the names of some CSS classes to improve readability
- Added some new CSS classes to _Master.css_, and implemented them in a few places
- Removed the light throbber, but kept the dark throbber, which is now just called _throbber.gif_. I realised that if we need a light throbber in the future, we can use a CSS filter to change the color of the remaining throbber.

## Videos

Before
Recording 2023-02-17 at 08 27 35

After
Recording 2023-02-17 at 08 53 48


iamlogand:
Any thoughts on this @tlscotarius?

tlscotarius:
I agree that the original dark colors are a little too harsh, but I'm not sure I like the current lighter colors.  I like the idea of a gradient, but not a fan of the dark red transitioning to a pink-ish color.  Maybe make the ending red of the gradient be darker (rather than pink).  Also, shift the start of the gradient further to the right so that all of the 'Republic of Rome Online" text is in the darker red.  

Also, I'm not a fan of the red for the title bar and then blue for the sign-in form.  Not sure if making the login form all red would be better though.  Might clash.  Probably good to look at some color.  Maybe worth going to a site like coolor.co to generate a color scheme and use the Tyrian Purple (#630330) as a starting point.

iamlogand:
> I agree that the original dark colors are a little too harsh, but I'm not sure I like the current lighter colors. I like the idea of a gradient, but not a fan of the dark red transitioning to a pink-ish color. Maybe make the ending red of the gradient be darker (rather than pink). Also, shift the start of the gradient further to the right so that all of the 'Republic of Rome Online" text is in the darker red.
> 
> Also, I'm not a fan of the red for the title bar and then blue for the sign-in form. Not sure if making the login form all red would be better though. Might clash. Probably good to look at some color. Maybe worth going to a site like coolor.co to generate a color scheme and use the Tyrian Purple (#630330) as a starting point.

You make good points (I wasn't that sure about these proposed colours either). I've marked this as draft as I'll need to make a few changes.

tlscotarius (dismissed):

tlscotarius (approved):

PR #123, opened 2023-06-18
Use websockets on the game page

### LINKED ISSUE #104: Use websockets on the game page
It would be useful if users could see live changes to the participant list whilst they are on the game page without having to refresh the page. Implement web sockets to meet this requirement.

iamlogand:
Closes #104

In addition to handling HTTP requests using Gunicorn with the root endpoint at `https://api.roronline.com/api/`, we now have the first WebSocket endpoint at `wss://api.roronline.com/ws/games/<id>` that uses Daphne and Channels.

The game lobby page uses a WebSocket connection to send a "status changed" message to other connected clients whenever a change is made to the game, such as:
- participant joins game
- participant leaves game
- game is deleted

When a client recieves a "status changed" message, it re-fetches the game, updating the page accordingly.

### Local development changes

For local development using the Django development server, create a Docker container with Redis, exposed on port 6379. This serves as the message layer for inter WebSocket consumer communication.

Install Docker, then run this command:
```
docker run -p 6379:6379 -d redis:5
```

Two new environment variables must be defined in the frontend .env.local file:

```
NEXT_PUBLIC_API_URL=http://127.0.0.1:8000/api/
NEXT_PUBLIC_WS_URL=ws://127.0.0.1:8000/ws/
```

The old variable called `NEXT_PUBLIC_BACKEND_ORIGIN` can be removed.


.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Git needs to allow shell scripts to have LF line endings
```
iamlogand, line 1: Test comment (ignore this)

PR #122, opened 2023-06-17
Created the form for join waitlist

SebastianHirsch:
Closes #110

A visitor can now join the waitlist by entering his email address in the respective form.

If the visitor enters a valid email address and submits the form, a popup window informs the visitor that he successfully joined the waitlist.

Join Waitlist Success

If the visitor enters an invalid email address, the helper text asks him to enter a valid one.

Join Waitlist Fail

PR #134, opened 2023-07-03
 User joining waitlist a second time should pretend first time join

### LINKED ISSUE #131: User joining waitlist a second time should pretend first time join

SebastianHirsch:
Closes #131 

The user gets a 201 Created HTTP response regardless whether the entered email already exists in the waitlist.


backend/rorapp/views/waitlist_entry.py
```diff
@@ -1,12 +1,24 @@
...
     queryset = WaitlistEntry.objects.all()
     serializer_class = WaitlistEntryCreateSerializer
     
+    def create(self, request, *args, **kwargs):
+        data = request.data
+
+        isEntryPresent = WaitlistEntry.objects.filter(email=data["email"])
+
+        if isEntryPresent:
+            return Response({}, status=status.HTTP_200_OK)
```
iamlogand, line 18: The problem with this solution is that a consumer of the API can still tell whether a new record has been created, or the record already exists. I appreciate that for anyone using the UI this solution is good, but it is also possible for anyone to send requests to the open API endpoint at https://api.roronline.com/api/waitlist_entry/ instead of using the UI, and if they do that they will see the status code in the response.

Instead of a 200, let's send a 201, so the API pretends to have created a new record:

```suggestion
            return Response({}, status=status.HTTP_201_CREATED)
```

frontend/pages/index.tsx
```diff
@@ -40,7 +40,7 @@ const HomePage = () => {
     const response = await requestWithoutAuthentication('POST', 'waitlist_entry/', { email });
 
     if (response) {
-      if (response.status === 201) {
+      if (response.status === 200 || response.status === 201) {
```
iamlogand, line 43: If you change the API as suggested then this change will not be necessary.

iamlogand (changes requested):
It looks like the "0013_alter_waitlistentry_options" migration should have been included in #122, but was forgotten. Oh well, good to include that here. The `verbose_name_plural` is a really minor thing anyway.

I think you're on the right track, just a couple things.

iamlogand:
Not essentail, but you could add the "closes" keyword to the description followed by the issue number and a sentence describing the change at a high level. See #133 for an example.

backend/rorapp/views/waitlist_entry.py
iamlogand: Thinking about it a bit more, I realise this wouldn't be enough. You will also need to make the actual response data (see the `{}` thing) look exactly like an authentic successul creation response!

SebastianHirsch: I added the needed data that should be returned.

frontend/pages/index.tsx
SebastianHirsch: I removed the 200 HTTP response condition.

SebastianHirsch:
> Not essential by any means, but you could add the "closes" keyword to the description followed by the issue number and a sentence describing the change at a high level. See #133 for an example.

I added it and hope it is as you wanted it.

backend/rorapp/views/waitlist_entry.py
```diff
@@ -20,7 +21,7 @@ def create(self, request, *args, **kwargs):
         isEntryPresent = WaitlistEntry.objects.filter(email=data["email"])
 
         if isEntryPresent:
-            return Response({}, status=status.HTTP_200_OK)
+            return Response({"email": data["email"], "entry_date": timezone.now()}, status=status.HTTP_201_CREATED)
```
iamlogand, line 24: This is perfect, well done! I tested it locally and the real/fake responses are indistinguishable.

iamlogand (approved):
Good job

PR #144, opened 2023-07-08
Changed styling of join waitlist to full width on mobile and fixed size for bigger screens

### LINKED ISSUE #140: Improve mobile responsiveness of the waitlist form
The form looks perfect on non-mobile displays, but could be improved on mobile:

image

SebastianHirsch:
Closes #140 

The join waitlist form takes full width for extra small screens. For bigger screens it is set to a fixed width of 380 px to align the style with the exclusive features buttons.

Join waitlist form responsivness

### DIFF
--- a/frontend/pages/index.tsx
+++ b/frontend/pages/index.tsx
@@ -82,13 +82,12 @@ const HomePage = () => {
 
       <h3 id="waitlist">Join our waitlist</h3>
         <form onSubmit={handleSubmit}>
-          <Stack alignItems="start" spacing={2}>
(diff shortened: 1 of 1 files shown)


iamlogand (changes requested):
You have certainly improved the appearance of the form on mobile, but you have also decided to change the way the button looks on larger screens by making the button wider for them too. I don't think it makes sense to have such a big button for desktop users. It was fine before, so I have added a suggested code change to resolve this.

Once that is fixed, I am happy to approve :)

iamlogand (approved):
Thanks 👍🏻

PR #367, opened 2024-01-12
Allow notification auto scroll to be switched off

### LINKED ISSUE #237: Sticky/free scroll modes for the notifications list
Currently, the notification list always scrolls down to the newest item whenever a new item is received. This can create a jarring UX if the user was trying read existing notifications.

To fix this, have the notification list track whether the user is at the bottom of the list. Whenever they are at the bottom, ensure that the current behavior of auto scrolling down keeps them at the bottom, so they see new notifications as they arrive. This I will call "sticky mode".

If, on the other hand, the user scrolls up, they exit sticky mode and go into "free mode". In free mode, it is assumed that they are browsing through existing notifications, so the auto scroll is disabled. They stay in free mode until they scroll back down to the latest notification at the bottom.

To help the user get back to the bottom and reenter sticky mode, there should be a floating button that appears whenever they are in free mode. Clicking this button will put the user back into sticky mode and scroll the user to the latest notification.

image

iamlogand:
Closes #237 by allowing the notification auto-scroll feature to temporarily switch off when the user is not already at the bottom of the list. In this case, there is now a button that can be clicked to scroll back to the bottom. The button is hidden when the user is at the bottom of the list.

image

### DIFF
--- a/frontend/components/GamePage.tsx
+++ b/frontend/components/GamePage.tsx
@@ -76,12 +76,11 @@ const GamePage = (props: GamePageProps) => {
     setAllTitles,
     setActionLogs,
     setSenatorActionLogs,
+    notifications,
+    setNotifications,
   } = useGameContext()
   const [latestActions, setLatestActions] = useState<Collection<Action>>(
     new Collection<Action>()
-  )
-  const [notifications, setNotifications] = useState<Collection<ActionLog>>(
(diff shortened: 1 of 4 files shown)


iamlogand:
Try to prevent the button from momentarily appearing when a new notification comes in

PR #386, opened 2024-01-27
Improve senator portrait image load time

iamlogand:
Attempt to improve senator portrait image load time by setting the images to `priority`.


iamlogand:
This didn't actually work

PR #686, opened 2026-06-19
feat: enable allied enthusiasm

iamlogand:
Now at the start of every initiative, there is a small chance of allied enthusiasm. If it happens twice in one turn, it becomes extreme allied enthusiasm. Either way, the State receives some talents (50 or 75) during the next revenue phase.

image

### Example logs

**During forum phase**
- _Faction 1 drew allied enthusiasm. With Rome's allies contributing additional funds, the State will receive 50T in the next revenue phase._
- _Faction 1 drew allied enthusiasm. With Rome's allies already enthusiastic, they are now extremely enthusiastic. The State will receive 75T in the next revenue phase._
- _Faction 1 drew allied enthusiasm. Rome's allies are already extremely enthusiastic so there is no additional effect._

**During revenue phase**
- _The State received an additional 50T due to allied enthusiasm._
- _The State received an additional 75T due to extreme allied enthusiasm._

### DIFF
--- a/backend/rorapp/effects/initiative_roll.py
+++ b/backend/rorapp/effects/initiative_roll.py
@@ -45,9 +45,7 @@ class InitiativeRollEffect(EffectBase):
             events = load_events()
             event_name = events["early_republic"].get(str(event_roll), "Unknown Event")
             if handle_event(game, current_faction, event_name):
-                game.sub_phase = Game.SubPhase.PERSUASION_ATTEMPT
-                game.save()
-            return True
+                return True
 
         next_card = game.draw_card()
         if next_card is not None:
--- a/backend/rorapp/effects/initiative_roll.py
+++ b/backend/rorapp/effects/initiative_roll.py
@@ -44,10 +44,10 @@ class InitiativeRollEffect(EffectBase):
             event_roll = random_resolver.roll_dice(count=3)
             events = load_events()
             event_name = events["early_republic"].get(str(event_roll), "Unknown Event")
-            if handle_event(game, game_id, current_faction.display_name, event_name):
+            if handle_event(game, current_faction, event_name):
                 game.sub_phase = Game.SubPhase.PERSUASION_ATTEMPT
                 game.save()
-                return True
+            return True
 
         next_card = game.draw_card()
         if next_card is not None:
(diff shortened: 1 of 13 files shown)


backend/rorapp/helpers/handle_event.py
```diff
@@ -0,0 +1,23 @@
...
+from rorapp.models import Game, Log
+
+
+def handle_event(
+    game: Game, game_id: int, faction_display_name: str, event_name: str
+) -> bool:
+    """Apply the event effect. Returns True if the event is implemented, False if not."""
+    if event_name == "Allied enthusiasm":
+        level = game.count_effect(GameEffect.ALLIED_ENTHUSIASM)
+        if level < 2:
```
paladin-devops: I think it's worth adding a log if the level is already at the max too, so that the players know there was an allied enthusiasm event, but there was no effect since it was at the max.

backend/rorapp/effects/initiative_roll.py
```diff
@@ -37,6 +39,16 @@ def execute(self, game_id: int, random_resolver: RandomResolver) -> bool:
         if not current_faction:
             return False
 
+        initiative = random_resolver.roll_dice(count=2)
+        if initiative == 7:
+            event_roll = random_resolver.roll_dice(count=3)
+            events = load_events()
+            event_name = events["early_republic"].get(str(event_roll), "Unknown Event")
```
paladin-devops, line 46: (non-blocking) Since only the early republic scenario is supported for now I think this is OK, but if it could be derived from the game state that would be great (for the future scenarios to be added).

iamlogand: Yeah, eventually I'll need to begin doing that.

For the time being I'll keep hard coding early republic scenario behaviors.

backend/rorapp/helpers/handle_event.py
iamlogand: Added new log for max enthusiasm via 48338e02e0cd10b5d8ab75e21588e963d69dfa09 and subsequent commits. Now whenever that card is drawn, the log created will always start with something like "Faction 1 drew allied enthusiasm..."

PR #689, opened 2026-06-21
fix: backend CI workflow_dispatch

iamlogand:
workflow_dispatch was broken since #682. This fixes it


iamlogand:
Proof it works now:

https://github.com/iamlogand/republic-of-rome-online/actions/runs/27902509011

PR #716, opened 2026-07-04
Award provinces on land victory & show Forum provinces in UI

paladin-devops:
## New Rule coverage
- 1.10.4: provinces created when war is eliminated by land victory
- 1.10.12: dual-strength wars require naval defeat before land battle; provinces awarded on the final land victory

### Backend
- `award_provinces_for_war` - looks up provinces by created_by in `province.json`, skips duplicates, handles compound strings ("Bequest or 2nd Cilician Pirates"), Illyrian wars gate (both Illyrian Wars), and Gallic war aliases
- Combat - calls award on war end only when the eliminating battle was a land victory (not naval_battle), per 1.10.4. Naval-only war endings and interim naval victories on dual-strength wars do not create provinces
- ProvinceSerializer - exposes computed frontier (from JSON) and in_forum (no governor yet)
- Test helper - POST /api/test/create-forum-provinces/<game_id>/ for E2E setup

### Frontend
- "Provinces in the Forum" panel in GameContainer.tsx — name, developed/undeveloped badge, frontier badge
- Province.ts types wired into PublicGameState

#### Screenshot
provinces-in-forum


frontend/components/GameContainer.tsx
```diff
@@ -622,6 +623,42 @@ const GameContainer = ({
               </>
             )}
 
+            {publicGameState.provinces.filter((p) => p.inForum).length >
+              0 && (
+              <>
+                <h3 className="mt-4 text-xl">Provinces in the Forum</h3>
```
iamlogand: **issue (blocking)**: an opinionated thing from me again: I don't think the "forum" as a concept should exist in this implementation of the game with respect to provinces because we can just label the province as "vacant", "governed by [senator]" or "in rebellion" instead, which is more intuitive and meaningful than "in the forum".

```suggestion
                <h3 className="mt-4 text-xl">Provinces</h3>
```

You might also avoid referring to "forum" in the codebase, but I'm not as fussed about this.

frontend/components/GameContainer.tsx
```diff
@@ -622,6 +623,42 @@ const GameContainer = ({
...
+                <div className="flex flex-col gap-4 lg:grid lg:grid-cols-[repeat(auto-fill,minmax(400px,1fr))]">
+                  {publicGameState.provinces
+                    .filter((p) => p.inForum)
```
iamlogand: **idea**: we could just show all provinces here regardless of whether they're vacant or governed. I think that'll be easier to design. Unless you had something more abitious in mind for displaying governed provinces?

backend/rorapp/helpers/provinces.py
```diff
@@ -0,0 +1,64 @@
...
+        if not _should_award_now(game, war, data["created_by"]):
+            continue
+
+        province = Province.objects.create(
+            game=game,
+            name=name,
+            developed=False,
+        )
+        created.append(province)
+        Log.create_object(game.id, f"{name} was established as a province.")
```
iamlogand, line 24: **praise**: concise, I like it

backend/rorapp/serializers/province.py
```diff
@@ -1,14 +1,26 @@
...
+    in_forum = serializers.SerializerMethodField()
+
     class Meta:
         model = Province
         fields = [
             "id",
             "game",
             "name",
             "developed",
+            "frontier",
```
iamlogand, line 14: **issue (blocking)**: I'm not a fan of frontier being calculated per serialization. It means the server would need to read data from a JSON to find out whether each province is a frontier every step.

How about calculating it once per Province instantiation and saving it to a new Province field called frontier?

I think saving it to DB once and reading the stored value each time would be faster. Open to debate on this - I may well be wrong.

backend/rorapp/serializers/province.py
```diff
@@ -1,14 +1,26 @@
...
+            "in_forum",
```
iamlogand: **issue (non-blocking)**: my preference would be to not pass in_forum over the API at all because the frontend will be able to easily determine whether there is a governor or not by reading `Province.governor` (a property that doesn't exist yet, but I imagine you will add a future PR when governor assignments are added).

Additionally I don't like "in forum" as a concept, as mentioned in my other comment.

backend/rorapp/helpers/provinces.py
iamlogand, line 1: **praise**: this core file looks good to go

iamlogand (changes requested):

frontend/components/GameContainer.tsx
paladin-devops: I'm good with displaying all provinces here! Will update.

backend/rorapp/serializers/province.py
paladin-devops: I almost added this as a new field on the model in my first pass, I mistakenly thought that reading from JSON was done for War as well, but I see now that the static data is added to the DB when the war is created. I will update this! I agree it would be faster.

iamlogand (approved):

PR #761, opened 2026-08-24
fix: return fleets to reserve after a fleet-only naval victory

### LINKED ISSUE #751: Naval victory with fleet-only force leaves fleets stranded and suppresses return log
When a fleet-only force wins a naval battle, the surviving fleets are left attached to a now-commanderless campaign instead of being returned to the active forces pool, and no log message is emitted for the commander's return to Rome.

ActualRory:
Fixes #751 -> When a force with no surviving legions won a naval battle that did not end the war, the commander was sent back to Rome but the surviving fleets stayed attached to the campaign, which was left commanderless.

Nothing was leaked permanently, the next senate phase eventually swept the campaign up. This will release the surviving fleets back to the reserve forces, delete the now-empty campaign, and log the return along with the reason for it. I reuse the phrasing the senate phase already used for the same cause, but that's easy to tweak if you fancy.

issue-751-3-with-fix


iamlogand (approved):
Log text looks good as is

Issue #762, opened 2026-08-24
Missing rule 1.10.5 implementation: Veteran Legion creation after land combat


SoyMakari:
Hi, 
I noticed this while testing a game: after winning a land battle with surviving Legions, no Legion was converted into a Veteran and no Allegiance was assigned to the Commander. Veteran Legion creation after land combat appears to be missing from the current combat resolution.

This appears to be a missing implementation of rule 1.10.5 (Veterans).

## Rule 1.10.5

Rule 1.10.5 states that one Legion among the survivors of a Land Battle resulting in:

- Stalemate
- Standoff
- Victory

is transformed into a Veteran Legion.

Veterans are not created in Naval Battles.

The corresponding Legion Allegiance marker is placed on the Commander's card until either:

- the Commander dies; or
- the Legion is eliminated.

The Legion itself remains Veteran even if it is later assigned to another Commander, and Veteran Legions count double in combat.

Rule 1.10.5 also contains additional Veteran creation rules for Civil War:

- one Senate Legion becomes Veteran following a Senate Victory;
- one Rebel Legion becomes Veteran following a Senate Defeat;
- one surviving Legion on each side becomes Veteran following a Civil War Stalemate.

The issue initially observed relates to normal Land Battles, but these Civil War cases may also need to be considered when implementing the complete rule.

## Selection of the new Veteran

The wording of rule 1.10.5 says "one Legion from among the survivors" but does not explicitly specify the selection method.

A reasonable implementation would be to randomly select one surviving Legion that is not already Veteran.

This would avoid selecting an already-Veteran Legion and would fit the existing use of controlled randomness in combat.

If this interpretation is adopted, it might also be worth recording it under the currently empty "Combat phase" section of `docs/rule-interpretations.md`.

If all surviving Legions are already Veterans, no additional Veteran would be created and no existing Allegiance should be reassigned.

## Current implementation

In:

`backend/rorapp/helpers/resolve_combat.py`

combat casualties are already resolved individually:

```python
destroyed_legions, surviving_legions = random_resolver.select_casualties(
    legions, legion_losses
)
Existing model support

Most of the required model support already exists.

In:

backend/rorapp/models/legion.py

the Legion model already contains:

veteran = models.BooleanField(default=False)

and:

allegiance = models.ForeignKey(
    Senator,
    related_name="legions_in_allegiance",
    blank=True,
    null=True,
    on_delete=models.CASCADE,
)

It also already implements Veteran combat strength:

@property
[...]

PR #769, opened 2026-08-29
feat: Veteran legions (§1.10.5)

ActualRory:
I love veteran legions, so I've taken a stab at #762.

This PR introduces the creation of Veteran legions off the back of a land battle. This is rule §1.10.5: One surviving non-veteran legion is promoted to Veteran after a land Victory, Stalemate or Standoff, and the Allegiance marker is placed on the commander.

### Behaviour

| Battle result | Veteran created? |
| --- | --- |
| Land Victory / Stalemate / Standoff | Yes, one survivor |
| Land Disaster / Defeat | No |
| Any naval battle | No |

- If the commander dies in that same battle the legion is still promoted, but owes Allegiance to nobody, since there is no card left to hold the marker.
- Already-veteran survivors are not re-promoted. If every survivor is already a Veteran, none is created.

### Allegiance lifecycle

- `Legion.allegiance` was `on_delete=CASCADE`, so deleting a senator would have deleted his veteran legions outright. Changed to `SET_NULL` (migration `0058`).
- `kill_senator` now clears allegiance explicitly, because family senators keep their row (as a dead senator, or as an heir) and would otherwise hold the marker forever. The legion stays a Veteran, it just loses the marker.

### Selection

Which survivor gets promoted is chosen at random in a real game: `RealRandomResolver.select_veteran` uses `random.choice`, and both `execute_effects_and_manage_actions` and `submit_action` default to that resolver.

The choice goes through `RandomResolver` alongside `select_casualties` so that tests can pin it. `FakeRandomResolver` sorts by an optional `veteran_order` list of legion names and returns the first, falling back to the lowest numbered legion when it is not set — unlike its other inputs (`dice_rolls`, `casualty_order`, `mortality_chits`), which raise when unset. That spares the 17 test files setting `casualty_order` and the 50 using the shared `resolver` fixture from edits to express a choice with no game consequence: legions are interchangeable, and the rule only says "one Legion from among the survivors".

### Frontend

- The deployment pools filtered on `allegiance === null`. That was dead code until now, but it would have made every Veteran undeployable the moment markers existed. §1.10.5 says a Veteran fights normally "even if placed under another Commander", so the filter is removed.
- Added a `pluralize` helper for the GameBar force counters, which otherwise read "1 veterans in reserve" now that exactly one Veteran per battle is the normal case.
- The GameBar counter is labelled "Veteran Legions". I'm not entirely happy with this, as it takes up a lot of space. Just plain 'Veterans' seems off. I'd like some other opinions on how to render this! I had a few options, will mull them over.

### Out of scope

Civil war Veterans (§1.11.371-373), as civil war battles are not implemented yet. I might take a stab at that later, but that's a pretty in-depth feature I haven't even looked at yet.

### Testing

- 11 new tests in `s1_10_5_veterans_test.py` covering each battle result, the naval exclusion, the already-veteran pool, resolver-driven selection, the dead-commander case, and allegiance release on death
- Full backend suite: 458 passed
- Playwright e2e: 7 passed (no longer gonna do these tests)
- Verified in the running app for both a land Victory and a Stalemate


ActualRory:
stalemate-02-veterans-popover

victory-02-veterans-popover

victory-03-log

Screenshots of the current version. As I say in main, I'm not satisfied with how we're rendering these in the topbar. Will have a think!

ActualRory:
Actually, should be solid as-is! 'Veteran Legions' text will be a placeholder for now. Someday, we'll have fancy icons.

backend/rorapp/helpers/resolve_combat.py
```diff
@@ -331,6 +331,25 @@ def resolve_combat(
...
+        promotable_legions = [l for l in surviving_legions if not l.veteran]
+        promoted_legion = random_resolver.select_veteran(promotable_legions)
+        if promoted_legion:
+            promoted_legion.veteran = True
+            # A dead commander has no card on which to place the allegiance marker
+            if not commander_killed:
+                promoted_legion.allegiance = commander
+            promoted_legion.save()
+            veteran_log_text = (
+                f"Legion {promoted_legion.name} was hardened into a Veteran Legion"
```
iamlogand: praise: I like the phrasing

backend/rorapp/helpers/resolve_combat.py
```diff
@@ -331,6 +331,25 @@ def resolve_combat(
...
+            )
+            if commander_killed:
+                veteran_log_text += ", owing allegiance to nobody."
```
iamlogand: nitpick: I would prefer the sentence to simply end with "...hardened into a Veteran Legion." in cases where the commander died in battle. I think "owing allegiance to nobody" feels slightly rebelious.

backend/rorapp/helpers/resolve_combat.py
```diff
@@ -331,6 +331,25 @@ def resolve_combat(
...
+            else:
+                veteran_log_text += f", owing allegiance to {commander.display_name}."
+            Log.create_object(game_id=game_id, text=veteran_log_text)
+
```
iamlogand, line 352: praise: Announcement order feels right

commander rewards/death -> veteran legions -> unit movements

FYI - in the future there is a good chance I'll introduce some kind of small delay to the issuance of logs, to increase the suspense. If I did that tomorrow, the current order would work. That's good.

frontend/helpers/text.ts
```diff
@@ -31,6 +31,13 @@ export function toSentenceCase(str: string): string {
   return str.charAt(0).toUpperCase() + str.slice(1)
 }
 
+/**
+ * Formats a count with its noun, pluralizing the noun when the count is not 1.
+ */
+export function pluralize(count: number, noun: string): string {
+  return `${count} ${noun}${count === 1 ? "" : "s"}`
+}
+
```
iamlogand, line 40: idea: Can't believe I never made this already! There are many places where it should be used, but that's for a separate issue

iamlogand (dismissed):
Looks great - thanks! The nitpick fix is optional

frontend/helpers/text.ts
iamlogand: (issue created)

backend/rorapp/helpers/resolve_combat.py
ActualRory: Oh man, that'd be a dope addition. Would make the game feel more alive.

backend/rorapp/helpers/resolve_combat.py
```diff
@@ -342,10 +342,10 @@ def resolve_combat(
                 promoted_legion.allegiance = commander
             promoted_legion.save()
             veteran_log_text = (
-                f"Legion {promoted_legion.name} was hardened into a Veteran Legion"
+                f"Legion {promoted_legion.name} hardened into a Veteran Legion"
             )
             if commander_killed:
-                veteran_log_text += ", owing allegiance to nobody."
+                veteran_log_text += ", serving the Republic with distinction."
```
ActualRory: @iamlogand Need a new approval, as I just commit this new wording.

frontend/helpers/text.ts
ActualRory: Figured we'd find a use for it elsewhere if I trojan-horsed it in here. I'll have a nosy at some point if it takes my fancy! I love a good refactor.

backend/rorapp/helpers/resolve_combat.py
ActualRory: Actually hang on - thought of something better.

backend/rorapp/helpers/resolve_combat.py
```diff
@@ -331,6 +331,25 @@ def resolve_combat(
...
+            promoted_legion.veteran = True
+            # A dead commander has no card on which to place the allegiance marker
+            if not commander_killed:
+                promoted_legion.allegiance = commander
+            promoted_legion.save()
+            veteran_log_text = (
+                f"Legion {promoted_legion.name} hardened into a Veteran Legion"
+            )
+            if commander_killed:
+                veteran_log_text += ", serving Rome with distinction."
```
ActualRory: @iamlogand How's this phrasing? Need another approval after the tweak commits.

ActualRory: I could alternatively just do a full stop here, thereby making it clinical when it doesn't have allegiance, so:

'Legion V hardened into a Veteran Legion, serving Rome with distinction.'
'Legion V hardened into a Veteran Legion, owing allegiance to Scipio Africanus.'

or

'Legion V hardened into a Veteran Legion.'
'Legion V hardened into a Veteran Legion, owing allegiance to Scipio Africanus.'

ActualRory:
I think it's beyond scope for this PR, but I'll need to rework the combat calculator to allow sending specific veteran legions - which is an important strategic decision.

(Screenshot from #778 where I stumbled across this.)
744-c-calculator-popover

backend/rorapp/helpers/resolve_combat.py
iamlogand: suggestion: could you drop the "serving Rome with distinction"? I'd prefer to simply end it with "...hardened into a Veteran Legion." in cases where the commander was killed.

The distinction idea is not bad, but doesn't really carry any information.

iamlogand (changes requested):

backend/rorapp/helpers/resolve_combat.py
ActualRory: Sure, was considering doing that one - will do it!

ActualRory: At my desk now, done!

iamlogand (approved):

PR #771, opened 2026-08-29
fix: broken "Forum - Attract Knight" preset

iamlogand:
The "Forum - Attract Knight" preset was broken: it wouldn't assign an initiative index, which is essential. I've also added talents to the faction taking the initiative, so they have some money available to spend.


iamlogand:
~TODO: test the other presets to see if they are also broken~

Edit: will investigate this separately -- #780

backend/rorapp/helpers/preset_loader.py
ActualRory, line 7: Dead import now that line 76 doesn't subscript the enum any more. Nothing will catch it either: no linter in `requirements.txt`, and mypy doesn't flag unused imports. Not blocking, but one we could do a broad sweep for these later in a refactor.

ActualRory:
Looks good regardless!

ActualRory (approved):

PR #772, opened 2026-08-29
feat: add the epidemic event (§1.07.21)

ActualRory:
Adds the **epidemic** event (a roll of 8 on the events table), implementing rule §1.07.21.

> Draw six Mortality Chits. Only those Senators currently in Rome can be killed.

### Behaviour

When a faction rolls 7 for initiative and then 8 on the events table:

- Six mortality chits are drawn through the existing draw logic, so "draw 2" chits and an exhausted bag behave exactly as they do in the mortality phase.
- Every living senator **in Rome** whose family code was drawn dies, unaligned senators in the forum included.
- Senators outside Rome (currently just commanders deployed to a war, but governors etc in future) are immune.
- As with the other events, no card is drawn and play moves straight on to persuasion attempts.

### New log examples

- _Faction 1 drew epidemic. A plague swept through Rome, killing 2 senators._
- _Faction 2 drew epidemic. A plague swept through Rome, killing a senator._
- _Faction 3 drew epidemic. A plague swept through Rome, but every senator in Rome survived._
- _Fabius of Faction 1 died in the epidemic._

`CauseOfDeath.EPIDEMIC` is new, so victims read as having "died in the epidemic" rather than of natural causes.

### Implementation notes

- `handle_event` now takes the `RandomResolver`. This is the first event needing randomness beyond the event roll itself.
- `handle_epidemic` reloads the game before returning. Deaths can release concessions back to the forum via `kill_senator`, which saves its own `Game` instance, and without the reload the sub-phase save in `handle_event` would clobber them.
- No new `GameEffect`. An epidemic leaves nothing behind, so there is nothing to render in the effects list and no frontend change. The deaths surface through the log.
- This touches `kill_senator.py`, as I'm also doing in #769, but in a different part of the file. The two merge cleanly in either order.

### Out of scope

Foreign epidemic, the dark blue side of this event. Foreign epidemic kills the first Governor, Proconsul, Captive or Rebel drawn who is away from Rome, and of those four only proconsuls exist so far. Repeat draws currently resolve as another epidemic. Added to the README checklist as a sub-item so we don't forget to add that.

### Testing

- 5 new tests in `s1_07_21_events_test.py`: a drawn senator dies, several drawn senators die, a senator away from Rome survives, exactly six chits are drawn (the sixth kills, the seventh is never reached), and no matches kills nobody
- Full backend suite: 452 passed
- mypy: clean on 284 source files
- Verified in the running app, forcing the event with a scripted resolver

### A good day for Faction 3

epidemic-event

### DIFF
--- a/backend/rorapp/helpers/handle_event.py
+++ b/backend/rorapp/helpers/handle_event.py
@@ -134,20 +134,15 @@ def handle_epidemic(
     ]
 
     prefix = f"{current_faction.display_name} drew epidemic."
-    if len(victims) >= 2:
+    if victims:
         Log.create_object(
             game.id,
-            f"{prefix} A plague swept through Rome, killing {len(victims)} senators.",
-        )
-    elif victims:
-        Log.create_object(
-            game.id,
-            f"{prefix} A plague swept through Rome, killing a senator.",
+            f"{prefix} A plague swept through Rome.",
         )
     else:
         Log.create_object(
             game.id,
-            f"{prefix} A plague swept through Rome, but every senator in Rome survived.",
+            f"{prefix} A plague swept through Rome, but every senator survived.",
         )
 
     for victim in victims:
--- a/backend/rorapp/helpers/handle_event.py
+++ b/backend/rorapp/helpers/handle_event.py
@@ -1,13 +1,23 @@
 from rorapp.classes.game_effect_item import GameEffect
-from rorapp.models import Faction, Game, Log
+from rorapp.classes.random_resolver import RandomResolver
+from rorapp.helpers.game_data import get_senator_codes
+from rorapp.helpers.kill_senator import CauseOfDeath, kill_senator
+from rorapp.models import Faction, Game, Log, Senator
 
 
-def handle_event(game: Game, current_faction: Faction, event_name: str) -> bool:
+def handle_event(
+    game: Game,
+    current_faction: Faction,
+    event_name: str,
+    random_resolver: RandomResolver,
+) -> bool:
     """Apply the event effect. Returns True if the event is implemented, False if not."""
     if event_name == "Allied enthusiasm":
         advances = handle_allied_enthusiasm(game, current_faction)
     elif event_name == "Drought":
         advances = handle_drought(game, current_faction)
+    elif event_name == "Epidemic":
+        advances = handle_epidemic(game, current_faction, random_resolver)
     elif event_name == "Evil Omens":
         advances = handle_evil_omens(game, current_faction)
     elif event_name == "Manpower Shortage":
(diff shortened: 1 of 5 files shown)


iamlogand:
This is the sort of feature that'll be fun to try out with the tool I'll be introducing via #773

backend/rorapp/helpers/handle_event.py
```diff
@@ -110,3 +120,39 @@ def handle_allied_enthusiasm(game: Game, current_faction: Faction) -> bool:
...
+    codes = set(random_resolver.draw_mortality_chits(6))
+    senators = Senator.objects.filter(game=game.id, alive=True)
+    victims = [
+        s
+        for s in senators
+        if s.location == "Rome" and get_senator_codes(s.code)[0] in codes
+    ]
+
+    prefix = f"{current_faction.display_name} drew epidemic."
+    if len(victims) >= 2:
```
iamlogand: suggestion: could you drop the "killing N senators"/"killing a senator" entirely?

backend/rorapp/helpers/handle_event.py
```diff
@@ -110,3 +120,39 @@ def handle_allied_enthusiasm(game: Game, current_faction: Faction) -> bool:
...
+        )
+    elif victims:
+        Log.create_object(
+            game.id,
+            f"{prefix} A plague swept through Rome, killing a senator.",
+        )
+    else:
+        Log.create_object(
+            game.id,
+            f"{prefix} A plague swept through Rome, but every senator in Rome survived.",
```
iamlogand: suggestion: could you remove "in Rome"? I think the first part of the sentence makes it clear enough we're talking about Rome

backend/rorapp/helpers/handle_event.py
```diff
@@ -110,3 +120,39 @@ def handle_allied_enthusiasm(game: Game, current_faction: Faction) -> bool:
...
+        )
+
+    for victim in victims:
+        kill_senator(victim, CauseOfDeath.EPIDEMIC)
+
+    # Reload the game, since deaths may have released concessions to the forum
+    game.refresh_from_db()
```
iamlogand, line 152: praise: good catch, I'd have missed this for sure

iamlogand (changes requested):
Players are going to love this feature!

backend/rorapp/helpers/handle_event.py
ActualRory: Yep! Good call

ActualRory: I appreciate it when you catch awkward wording like this!

backend/rorapp/helpers/handle_event.py
ActualRory: Done!

iamlogand (approved):

iamlogand:
`not-a-covid-reference` 😆

PR #773, opened 2026-08-29
feat: add a fake random resolver to the UI as a debug-only tool

iamlogand:
Adds a Debug Tools panel for overriding dice rolls, casualty order, and mortality chit outcomes through the UI, with a live queue display and per-type controls. Enabled only when the TEST_ENDPOINTS_ENABLED Django setting is True.

The resolver queue is stored in cache and consumed one value per call, so tests that set resolver values must queue them in the correct order. Existing tests have been updated to use this pattern.

The fake resolver for casualties and veteran selection now accepts Arabic unit numbers rather than Roman numerals.

## Example

Here's an example where I guaranteed a successful knight attraction for Faction 1 (dice roll 6), then a random event for the next faction (dice roll 7), then an epidemic (dice roll 8), and finally that the victims will be Cornelius and Claudius (mortality chits 1 and 5).

image

Moments later...

image


iamlogand:
Will wait for #769 to hit main, then add UI random resolver support for veteran selection

ActualRory:
This is dope! Will be super handy

ActualRory (approved):
Everything looks grand to me!

PR #776, opened 2026-08-29
feat: eliminate forces during the senate phase

ActualRory:
Implements the second half of rule 1.09.63 - the Senate may vote to raise **or eliminate** forces. This is as a new "Propose eliminating forces" senate proposal, sitting alongside the existing "Propose raising forces". Ticks `Eliminate forces` off the readme.

**Edit**: I'm under some debate about whether eliminate is actually a good term here, see comments below.

## Rules

> The Senate may vote to raise or eliminate Forces. […] A Legion or Fleet cannot be both eliminated and built (or rebuilt) in the same Senate Phase. […] The Senate may only vote to eliminate units located in the Active Forces boxes. (1.09.63)

- **Reserve only** - only legions and fleets not attached to a campaign are eligible, enforced both in the form and in `execute()`.
- **No raise-then-eliminate** - units raised earlier in the same senate phase are excluded from the form and rejected by `execute()`, via new `recently_raised` flags on `Legion` and `Fleet`.
- **No eliminate-then-rebuild** - the numbers of eliminated units are held on the game, so raising forces skips them and the force pool cap accounts for them. Both restrictions are cleared when the senate phase ends.

## Changes

**Backend**

- `actions/propose_eliminating_forces.py` - new `ProposeEliminatingForcesAction`, offering a multiselect of eligible legions and one of eligible fleets.
- `effects/proposal_eliminate_forces.py` - new `ProposalEliminateForcesEffect`, resolving the vote and removing the units.
- `models/` - `recently_raised` on `Legion` and `Fleet`, `eliminated_legion_numbers` and `eliminated_fleet_numbers` on `Game` (migration `0058_eliminate_forces`).
- `effects/proposal_raise_forces.py` and `actions/propose_raising_forces.py` - skip eliminated numbers when assigning numbers to new units and when computing the recruitment cap.
- `effects/senate_phase_end.py` - reset both restrictions when the senate closes.
- `helpers/unit_lists.py` - extracted `string_to_unit_lists` from the recall effect, now shared by both proposals.
- Proposal `POSITION` values shifted by one so the new action sits next to raising forces.

**Frontend**

- `ActionDescription.tsx` - a description for the new proposal. Nothing else needed, as the generic action form already renders the multiselects.

## Tests

`tests/s1_basic_game/s1_09_senate_phase/s1_09_63_eliminate_forces_test.py` covers the passed and defeated motion, the reserve-only and recently-raised restrictions in both the schema and `execute()`, the proposal text, the eliminate-then-rebuild block, and the reset at the end of the senate phase.

`mypy` is clean and the full backend suite passes (456 tests).

## Verified in the app

Played it through in a 3-player game: proposed eliminating 3 legions and 1 fleet, voted it through, and confirmed the reserve counts dropped from 6 and 4 to 3 and 3 with logging. A follow-up proposal to raise 3 legions produced legions VII-IX rather than reusing I-III, and those newly raised legions were not offered for elimination.

## Budget cuts in motion
Probably gonna tweak the exact wording and not bother to update these screenshots, but you get the jist!

02-proposal-form
03-motion-on-the-floor
04-forces-eliminated
05-raised-forces-not-eligible


ActualRory:
Honestly I don't love the word 'Eliminate' for this action, any issue with me adding this to [terminology-changes.md](https://github.com/iamlogand/republic-of-rome-online/blob/main/docs/terminology-changes.md)? It's eliminate in the old rulebook, but I feel like disband is much more accurate (you're not killing the legion) in the way of the extant drought vs famine change. Pushback is fine here, if you disagree!

iamlogand:
I agree that "eliminate" sounds a bit off. Your suggestion to use "disband" instead sounds like a good fit.

docs/terminology-changes.md
```diff
@@ -9,3 +9,7 @@ The term "below minimum force" has been replaced with "risky command". Since pla
 ## Drought → Famine
 
 The term "drought" has been replaced with "famine". Droughts are natural occurrences that aren't caused by humans. However, in this game, warfare can trigger a drought. Famines are more fitting since they were often caused by droughts or war.
+
+## Eliminate forces → Disband forces
+
+The term "eliminate forces" has been replaced with "disband forces". The senate can only vote to stand down legions and fleets that are sitting in the reserve, so "disband" describes what happens more plainly than "eliminate", which suggests they were destroyed by an enemy. Note that the original game also uses "disbandment" for the separate advanced rule that lets a victorious commander stand down his own veteran legions ([2.04.2](../game-data/rules/02-advanced-game/2.04-legionary-disbandment-recall.md#2.04.2)). I further propose we use "retire" instead of "disbandment" for that advanced rule, as that's much more accurate there too.
```
iamlogand: suggestion: I also like your idea to change OG "disbandment" to "retire". Can you split this off into it's own dedicated `## Disbandment → Retire` item in this document?

For future, I think the best place to track proposals like this would be GitHub issues rather the documentation itself

backend/rorapp/effects/proposal_raise_forces.py
```diff
@@ -54,19 +54,30 @@ def execute(self, game_id: int, random_resolver: RandomResolver) -> bool:
...
+            ] + game.disbanded_fleet_numbers
 
             new_legions: List[Legion] = []
             new_fleets: List[Fleet] = []
             for num in range(1, 26):
                 if legions_to_raise and num not in unavailable_legion_nums:
-                    new_legions.append(Legion.objects.create(game=game, number=num))
+                    new_legions.append(
+                        Legion.objects.create(
+                            game=game, number=num, recently_raised=True
```
iamlogand: suggestion (non-blocking): instead of explicitly setting `recently_raised` to True here, you could set the default value for `recently_raised` to true, since we'll always want this to be true when creating new legions.

backend/rorapp/models/game.py
```diff
@@ -256,3 +258,21 @@ def clear_unavailable_proposals(self) -> None:
     def clear_senate_sub_phase_proposals(self) -> None:
         self.clear_defeated_proposals()
         self.clear_unavailable_proposals()
+
+    # disbanded unit number methods
+
+    # A legion or fleet cannot be both disbanded and raised in the same senate
+    # phase (1.09.63), so disbanded numbers stay out of the force pool until
+    # the senate phase ends.
```
iamlogand: issue: Could you remove these three comment lines? I'd like to keep code comments to a minimal here

iamlogand (changes requested):
Solid PR overall, just a couple small things

backend/rorapp/effects/proposal_raise_forces.py
ActualRory: Good call! That'd be tidier

backend/rorapp/models/game.py
ActualRory: Sure!

docs/terminology-changes.md
ActualRory: Done! I've sorted them alphabetically, and noted for in future!

backend/rorapp/effects/proposal_raise_forces.py
ActualRory: Did some refactoring to this end, see commit [b3f4ffd](https://github.com/iamlogand/republic-of-rome-online/pull/776/commits/b3f4ffdb02a89cc93a097e8734776322e323fffe) and let me know what you think!

iamlogand (dismissed):

ActualRory:
All hail the implementation of budget cuts. I'm sure Rome won't need those legions later.

ActualRory:
Needed a migration renumber to pass checks, no actual changes but need another approval! @iamlogand

iamlogand (approved):

PR #778, opened 2026-08-30
feat: show base strength and individual modifiers for wars (§1.07.332, §1.07.342)

### LINKED ISSUE #744: For modified war strengths, show the base and individual modifiers
Currently when a war is matched or has an enemy leader, the strength is increased, but the UI doesn't show why. It just shows the new number. It's unclear what the base strength is and what specific things are modifying it.

ActualRory:
Addresses #744.

When a war is multiplied by Matching Wars or joined by an enemy Leader, the UI showed only the resulting number, with nothing to say where it came from.

> **§1.07.332 MATCHING WARS** - Whenever two Wars of the same type are active (e.g., two Punic Wars or two Cilician Pirates), the Land and Fleet Strength (but not Fleet Support) of each are doubled as long as the two remain active. If a third Matching War is active, the Strength numbers of all three are tripled. If all four Macedonian War cards are active at the same time, their Strength is quadrupled.

> **§1.07.342 LEADER STRENGTH** - While a Leader is active and matched with the War(s), the Land and Fleet Strength (not Fleet Support) number(s) of the War(s) are increased by his value, after any relevant doubling, tripling, or quadrupling for Matching Wars.

### Behaviour

A modified Land or Naval strength is now a hover/focus target that opens a breakdown naming every contributor: the base, the matching wars driving the multiplier, and each active leader. Unmodified strengths render as before, as a plain number with no hover. If you reckon I should include the hover there too, I can do that - but I figured it'd waste people's time hovering over it.

The same breakdown appears wherever the modified figure is shown: on the war cards and beside `War strength` in the combat calculator. Fleet Support is never modified, so it is left alone.

### Implementation notes

The `base × matchingWarMultiplier + leaderStrength` calculation was duplicated between `GameMain` and `CombatCalculatorItem`. It now lives in one place, `getWarStrengthBreakdown` in `frontend/helpers/wars.ts`, which returns the total along with the parts that produced it; `WarStrength` renders that through the existing `Popover`. No backend changes there, the serializer already sends the unmodified strengths.

Caught a couple bugs while putting this one together:
- **Inactive leaders were being counted.** Both call sites filtered enemy leaders on `series_name` alone, ignoring `active`. A leader drawn while no matching war is active, or one that withdrew after a Roman victory, was inflating the displayed strength even though `resolve_combat` ignores it. With an inactive Hamilcar (Punic, +3) in the Forum, the 1st Punic War now shows 27 rather than 30. The same filter now also feeds the calculator's disaster/standoff probabilities, which had the same problem. I fixed it here, and as such you will see me fumble with the labels for this PR.
- **Popovers were painting behind the combat calculator.** The calculator sits at `z-index: 1000` and `Popover` portalled at `z-50`, so a breakdown opened inside the calculator was invisible. The popover portal now sits above it.

744-b-land-strength-popover
744-c-calculator-popover

### Testing

- `npx tsc --noEmit` and `eslint` clean on the changed files. (`npm run lint` is broken repo-wide on `main` — `next lint` was removed in Next 16 - so eslint was run directly.)
- Backend suite unaffected and still green: **447 passed** (`pytest`).
- Checked in a test game with two active Punic Wars, an active Hannibal (+7) and an inactive Hamilcar (+3): 1st Punic land 10×2+7 = 27, naval 10×2+7 = 27, fleet support unchanged at 5; 2nd Punic land 15×2+7 = 37; 1st Gallic 10 with no popover.

### DIFF
--- a/frontend/components/CombatCalculatorItem.tsx
+++ b/frontend/components/CombatCalculatorItem.tsx
@@ -1,14 +1,18 @@
 import { useCallback } from "react"
 
 import CombatCalculation from "@/classes/CombatCalculation"
-import EnemyLeader from "@/classes/EnemyLeader"
 import PublicGameState from "@/classes/PublicGameState"
-import { getEvilOmensLevel } from "@/helpers/gameEffects"
 import Senator from "@/classes/Senator"
 import War from "@/classes/War"
-import getDiceProbability from "@/helpers/dice"
+import WarStrength from "@/components/WarStrength"
 import { SERIES_NULLIFIERS } from "@/data/statesmen"
-import { compareWars } from "@/helpers/wars"
+import getDiceProbability from "@/helpers/dice"
+import { getEvilOmensLevel } from "@/helpers/gameEffects"
+import {
+  compareWars,
+  getActiveLeaders,
+  getWarStrengthBreakdown,
+} from "@/helpers/wars"
 
 interface CombatCalculatorItemProps {
   publicGameState: PublicGameState
@@ -45,9 +49,9 @@ const CombatCalculatorItem = ({
     (war: War) => war.id === combatCalculation.war,
   )
 
-  const enemyLeaders = publicGameState.enemyLeaders.filter(
-    (leader: EnemyLeader) => leader.seriesName === war?.seriesName,
-  )
+  const enemyLeaders = war
+    ? getActiveLeaders(war, publicGameState.enemyLeaders)
+    : []
 
   const setWar = (warId: number | null) => {
     if (combatCalculation.war !== warId) {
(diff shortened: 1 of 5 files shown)


iamlogand:
A question on inactive leaders being counted -- is it possible within the rules for an enemy leader to be inactive whilst a matching war exists?

iamlogand (approved):
Looks great! Cheers

ActualRory:
> A question on inactive leaders being counted -- is it possible within the rules for an enemy leader to be inactive whilst a matching war exists?

There's exactly one unlikely and niche scenario for this. 

Per 1.07.3321, Leaders do not activate Imminent Wars. While a matching War sits in the Imminent Wars box (which lasts only until the start of the next turn, or until the Senate attacks it), it's still treated as Inactive. So if a Leader is drawn (or is already in the Curia) and his only matching War in play is currently in the Imminent Wars space, that Leader stays inactive in the Curia until that war is active.

PR #782, opened 2026-08-30
feat: Consul for Life (§1.09.82)

ActualRory:
Implements the Consul for Life mechanic: election (§1.09.821), automatic appointment (§1.09.822), and the victory it confers (§1.12.2). These were the last three unchecked Senate and Revolution-phase items on the README checklist, though I have not tackled the faction takeover mechanic of this yet, that'll be fiddly and this is still useful without it.

> **1.09.821 CONSUL FOR LIFE (ELECTION)**
> At any time during the Senate Phase following the Consular elections, the Presiding Magistrate, or a player using a Tribune, may nominate any Aligned Senator in Rome with at least 21 Influence "Consul for Life". The exception to this is it cannot be proposed while the Censor is Presiding Magistrate during Prosecutions. The nominee may already hold an office. For this vote, the nominee may add his Influence to his vote total. Consul for Life can be proposed only once per turn and cannot be vetoed. [...] Assassination attempts may take place during such a vote, but once elected or appointed, the Consul For Life cannot be assassinated [...]

> **1.09.822 CONSUL FOR LIFE (APPOINTMENT)**
> If at any time in the Game an Aligned Senator is in Rome and has at least 35 Influence, and the Military Reforms Law (Middle Republic) is not already in place, he immediately becomes the Consul for Life. If two or more Senators qualify in the same turn, the one exceeding 34 by the most is the Consul for Life. If equal, the Senator belonging to the Faction with the most combined Influence is Consul for Life. Achieving Consul for Life in this manner is an automatic appointment - there is no chance for an Assassination attempt. In the exceedingly rare event that Senators are both elected and appointed Consul for Life in the same turn, the elected one ceases to be Consul for Life.

### Behaviour

- **Election.** The Presiding Magistrate, or a faction using a Tribune, may nominate any aligned senator in Rome with at least 21 influence. Available during the dictator election, censor election and other business — every post-consular sub-phase that accepts proposals, excluding prosecutions.
- **Nominee's influence counts as votes.** The senator named in the proposal adds his influence to whichever side he votes, across the simple, split and bought-vote paths.
- **Once per turn.** The lock is set when the motion reaches the floor, not when it resolves, so an assassinated nominee cannot be swapped for another that turn (§1.09.721). It clears at the start of the next Senate Phase.
- **Cannot be vetoed** (§1.09.152).
- **Automatic appointment.** Any aligned senator in Rome reaching 35 influence takes the title immediately, in any phase, with no vote. Ties break on influence, then on the faction's combined influence. An appointment supersedes an election made the same turn; an appointed Consul for Life is never displaced.
- **Assassination immunity.** The title-holder is removed from the target list and rejected server-side. A nominee mid-vote is still a legal target.
- **Not an office** (§1.09.823). It grants no influence and leaves the HRAO and Presiding Magistrate untouched. It also survives the next consular election, unlike Dictator or Censor.
- **Victory.** Alive at the end of a Revolution Phase, his faction wins. On Era Ends he takes precedence over the combined-influence count.

### New log examples

```
Cornelius of Faction 1 was elected Consul for Life.
Cornelius of Faction 1 reached 35 influence and became Consul for Life.
Game over! Cornelius ruled on as Consul for Life, so Faction 1 wins.
The era has ended! Cornelius ruled as Consul for Life, so Faction 1 wins.
```

### Implementation notes

- Mirrors the existing Dictator structure: `helpers/consul_for_life.py` holds eligibility, the vote bonus and the single `grant_consul_for_life` choke point; `NominateConsulForLifeAction` writes the `Elect Consul for Life X` proposal string and `ElectConsulForLifeEffect` resolves it on that prefix.
- `senate_open_for_proposals` now delegates to a new `senate_open_for_proposals_in`, since this is the first proposal valid in more than one sub-phase. The single-sub-phase signature is unchanged for every existing caller.
- The once-per-turn lock needed a new `Game` field rather than `unavailable_proposals`, which `clear_senate_sub_phase_proposals()` wipes on every sub-phase transition.
- `AppointConsulForLifeEffect` validates in any phase, so `Game.consul_for_life_appointed` is what stops the effect loop re-firing; it also encodes that an appointment supersedes an election but never the reverse.
- `GameOverConsulForLifeEffect` shares `RevolutionPhaseEndEffect`'s trigger and sits ahead of it in the registry.

### Out of scope

- **§1.09.823 faction takeover** - the Consul for Life's player should control all non-rebel factions during the Senate Phase. That needs changes to faction resolution in `views/submit_action.py`, per-player websocket state and the frontend action panel, so I'm leaving that for a follow-up. This is reflected in the project README.
- The Rebel carve-outs in §1.09.823 and the captive clause in §1.10.71, neither of which has supporting mechanics yet.
- **Military Reforms** (§1.09.822) should block the automatic appointment, but Laws are not yet implemented and aren't part of the early republic scenario - so that's a battle for another day.
- Intrigue cards cancelling a tribune-raised nomination (e.g. Graft), again, not part of the early republic scenario so no problems there yet.

### Testing

- `s1_09_821_consul_for_life_election_test.py` (18), `s1_09_822_consul_for_life_appointment_test.py` (10), `s1_12_2_consul_for_life_win_test.py` (4), plus 3 added to `s1_09_7_assassination_test.py`.
- Full backend suite: **493 passed** (458 on `upstream/main` before this branch). `mypy .`: **no issues in 411 source files**. `tsc --noEmit` and `next build` both clean.


ActualRory:
image

ActualRory:
01-nominate-consul-for-life
02-proposal-on-the-floor
03-advanced-vote-influence-counts
04-elected-consul-for-life
05-appointed-at-35-influence
06-consul-for-life-wins

ActualRory:
"Elect Consul for Life Cornelius" is a bit clunky, but I'm using the standard way and didn't really want a one-off.
Might make a follow-up PR to change all election motion messages to a"Elect Cornelius as Consul for Life" format.

backend/rorapp/actions/nominate_consul_for_life.py
```diff
@@ -0,0 +1,99 @@
...
+from rorapp.classes.random_resolver import RandomResolver
+from rorapp.game_state.game_state_live import GameStateLive
+from rorapp.game_state.game_state_snapshot import GameStateSnapshot
+from rorapp.helpers.consul_for_life import (
+    CONSUL_FOR_LIFE_PREFIX,
+    get_eligible_consul_for_life_candidates,
+)
+from rorapp.helpers.proposal_available import consul_for_life_proposal_available
+from rorapp.helpers.senate_proposal import (
+    CONSUL_FOR_LIFE_SUB_PHASES,
```
iamlogand: issue (blocking): Why does `CONSUL_FOR_LIFE_SUB_PHASES` ever leave senate_proposal.py?

It's only used once -- as an argument when you call `senate_open_for_proposals_in()`. Both `senate_open_for_proposals_in()` and `CONSUL_FOR_LIFE_SUB_PHASES` are defined in senate_proposal.py. So you could have just used `CONSUL_FOR_LIFE_SUB_PHASES` directly in the `senate_open_for_proposals_in()` definition.

backend/rorapp/helpers/senate_proposal.py
```diff
@@ -16,32 +17,50 @@
...
+        and game_state.game.sub_phase in sub_phases
         and (
             game_state.game.current_proposal is None
             or game_state.game.current_proposal == ""
         )
     )
 
 
+def senate_open_for_proposals(game_state, sub_phase) -> bool:
+    return senate_open_for_proposals_in(game_state, (sub_phase,))
```
iamlogand: question: Why do we need two functions so similarly named (`senate_open_for_proposals_in` and `senate_open_for_proposals`), where one of them just calls the other? I'm a bit confused by this.

backend/rorapp/models/game.py
```diff
@@ -109,11 +113,15 @@ def status(self) -> str:
 
     @property
     def votes_pending(self: "Game") -> int:
+        from rorapp.helpers.consul_for_life import consul_for_life_vote_bonus
```
iamlogand, line 117: question: Why use an inline import rather than one at the top?

backend/rorapp/effects/senate_phase_start.py
```diff
@@ -17,6 +17,8 @@ def validate(self, game_state: GameStateSnapshot) -> bool:
     def execute(self, game_id: int, random_resolver: RandomResolver) -> bool:
 
         game = Game.objects.get(id=game_id)
+        game.consul_for_life_proposed = False
```
iamlogand: suggestion: You could move this to `SenatePhaseEndEffect`, since we already perform senate phase cleanup things there (unless there's a good reason you put it here).

backend/rorapp/effects/appoint_consul_for_life.py
```diff
@@ -0,0 +1,23 @@
...
+
+class AppointConsulForLifeEffect(EffectBase):
+
+    def validate(self, game_state: GameStateSnapshot) -> bool:
+        if game_state.game.consul_for_life_appointed:
+            return False
+        return get_consul_for_life_appointee(game_state.senators) is not None
+
+    def execute(self, game_id: int, random_resolver: RandomResolver) -> bool:
+        snapshot = GameStateSnapshot(game_id)
```
iamlogand: question: Why did you use a game state snapshot here?

backend/rorapp/effects/game_over_consul_for_life.py
```diff
@@ -0,0 +1,40 @@
...
+
+    # The Consul for Life wins if he survives to the end of a Revolution Phase (1.12.2)
+    def validate(self, game_state: GameStateSnapshot) -> bool:
+        return (
+            game_state.game.phase == Game.Phase.REVOLUTION
+            and game_state.game.sub_phase == Game.SubPhase.PLAY_STATESMEN_CONCESSIONS
+            and all(
+                f.has_status_item(FactionStatusItem.DONE) for f in game_state.factions
+            )
+            and get_consul_for_life(game_state.senators) is not None
```
iamlogand, line 21: nitpick: the `is not None` is redundant

backend/rorapp/helpers/proposal_available.py
```diff
@@ -24,6 +28,21 @@ def dictator_election_proposal_available(game_state) -> bool:
     return censor_election_proposal_available(game_state)
 
 
+def consul_for_life_proposal_available(game_state) -> bool:
+    game = game_state.game
+    # Consul for Life can be proposed only once per turn (1.09.82)
+    if game.consul_for_life_proposed:
+        return False
+    if get_consul_for_life(game_state.senators) is not None:
```
iamlogand, line 35: nitpick: the `is not None` is redundant

backend/rorapp/helpers/proposal_available.py
```diff
@@ -24,6 +28,21 @@ def dictator_election_proposal_available(game_state) -> bool:
     return censor_election_proposal_available(game_state)
 
 
+def consul_for_life_proposal_available(game_state) -> bool:
+    game = game_state.game
+    # Consul for Life can be proposed only once per turn (1.09.82)
```
iamlogand: nitpick: I think this comment is unnecessary

backend/rorapp/actions/vote_yea.py
iamlogand, line 1: idea (non-blocking): It would be nice if the log for when the nominee votes yea in a consul for life election had some indication that the nominee got N additional votes due to his influence.

backend/rorapp/helpers/consul_for_life.py
```diff
@@ -0,0 +1,99 @@
...
+    previous_holders = Senator.objects.filter(
+        game=game_id, titles__contains=[Senator.Title.CONSUL_FOR_LIFE.value]
+    )
+    for previous_holder in previous_holders:
+        if previous_holder.id != senator_id:
+            previous_holder.remove_title(Senator.Title.CONSUL_FOR_LIFE)
+            previous_holder.save()
+
+    # Consul for Life is not an office, so it grants no influence and does not
+    # affect the HRAO or presiding magistrate (1.09.823)
```
iamlogand: nitpick: I don't think these comments are necessary

backend/rorapp/models/game.py
```diff
@@ -93,6 +93,10 @@ class SubPhase(models.TextChoices):
     assassination_roll_modifier = models.IntegerField(default=0)
     assassination_roll_result = models.IntegerField(default=0)
     bodyguard_rerolls_remaining = models.IntegerField(default=0)
+    # Consul for Life may be nominated only once per turn (1.09.82)
+    consul_for_life_proposed = models.BooleanField(default=False)
+    # Set once a Consul for Life has been automatically appointed (1.09.822)
+    consul_for_life_appointed = models.BooleanField(default=False)
```
iamlogand, line 100: prasie: It took me a long time to understand why this field was necessary! But given rule 1.09.822 I can see why you had to add something like this

backend/rorapp/models/game.py
```diff
@@ -93,6 +93,10 @@ class SubPhase(models.TextChoices):
     assassination_roll_modifier = models.IntegerField(default=0)
     assassination_roll_result = models.IntegerField(default=0)
     bodyguard_rerolls_remaining = models.IntegerField(default=0)
+    # Consul for Life may be nominated only once per turn (1.09.82)
+    consul_for_life_proposed = models.BooleanField(default=False)
```
iamlogand, line 98: issue (blocking): this is redundant. Instead of adding this, you could check `game.defeated_proposals` and call `get_consul_for_life()` to find out whether a consul for life proposal has been added.

With models I generally prefer to avoid duplicating information, even if there is a small performance cost as a result.

iamlogand (changes requested):
Thanks for developing this feature!

I did find some issues. None are that bad (I didn't find anything that looked like a bug). I'm just holding you to a high standard.

I've not tested this locally, but will do so after the comments have been addressed.

backend/rorapp/helpers/senate_proposal.py
ActualRory: Confused myself, this is a Claude artifact. Seems like a extra function was stood up instead of widening the existing one. Bad QC on my part, sorry! Should have caught this. I've rewrote this as a widened single function.

backend/rorapp/helpers/consul_for_life.py
ActualRory: Done!

backend/rorapp/models/game.py
ActualRory: Can I? Looking at game.defeated_proposals, it seems to only record if it goes through. An assassinated nominee (and probably a tribune veto) cancels the proposal via `clear_proposal_state()` and nothing gets recorded in there. I could alternatively flesh out `defeated_proposals` to cover cancelled ones with a third status - but I saw that as a bit more invasive. Let me know if I'm right there and what you think!

backend/rorapp/helpers/proposal_available.py
ActualRory: Done!

backend/rorapp/effects/game_over_consul_for_life.py
ActualRory: Done!

backend/rorapp/effects/appoint_consul_for_life.py
ActualRory: Mistake carried over from staring at `validate`, dumb! Now fixed, the snapshot alone cost as much as the entire rest of execute so this was very daft.

backend/rorapp/models/game.py
ActualRory: Yeah I've been looking at this more, that list is sub-phase-scoped while the restriction is turn-scoped, and the assassinated nominee thing is a snag

backend/rorapp/actions/vote_yea.py
ActualRory: Good idea! Added log_consul_for_life_vote_bonus() and called it from all three vote actions.

backend/rorapp/effects/senate_phase_start.py
ActualRory: Good suggestion! No good reason, I've moved it in 9d69dd6.

backend/rorapp/models/game.py
ActualRory: If I do it at the top, Django won't play ball on start. `game.py `gets pulled in from the first line of `rorapp/models/__init__.py `(available_action → faction → game), so a top-level import of the helper runs while `rorapp.models` is still empty. 

The helpers from rorapp.models import then fails:
`ImportError: cannot import name 'Game' from partially initialized module 'rorapp.models'`

Importing submodules directly hits the same wall, since `models/senator.py` imports `game.py `too.

military_crisis a few lines below already does this with from rorapp.models.war import War (since #662). I think you ran into the same snag back when making that, these are the only two deferred imports in models/, both for this reason.

ActualRory: Past my bedtime, hope that makes sense!

backend/rorapp/helpers/proposal_available.py
ActualRory: Fixed, you're right!

backend/rorapp/actions/nominate_consul_for_life.py
ActualRory: Shouldn't have been exported!

My other commits bulldozed through this one a bit, and the wider shape is different now: 7678652 removed the senate_open_for_proposals_in() wrapper and made senate_open_for_proposals() accept a variable number of arguments, so there's no longer a two-function split to hide the constant inside. That left CONSUL_FOR_LIFE_SUB_PHASES with a single consumer and no reason to exist, so 06b3de7 deletes it and passes the three sub-phases inline, matching the other 14 call sites. The rule comment is still knocking around in there but will have a nosy tomorrow.

If you'd still prefer it encapsulated, I can add a consul_for_life_open_for_proposals(game_state) predicate in senate_proposal.py

ActualRory:
Need to re-test this one myself too (in light of a soup of new commits) feel free to test yourself in the meantime

backend/rorapp/models/game.py
iamlogand: Good point! You were right to add this field, it's not redundant

backend/rorapp/models/game.py
iamlogand: Makes perfect sense, thanks for explaining!

backend/rorapp/actions/nominate_consul_for_life.py
iamlogand: You've cleaned this up a lot since I last saw it, so happy to mark this resolved

backend/rorapp/effects/roll_assassination_dice.py
iamlogand, line 1: note: your change here is good. You'll probably see many places where I fetch a related entity in a weird way like this:

```
target_faction = Faction.objects.get(game=game_id, id=target.faction_id)
```

...when I could have easily used Django's shortcut:

```
target_faction = target.faction
```

It was a bad habbit, and there was nobody else to let me know what I was doing wrong. I don't use Django at work so I've been in a bit of a bubble. If you see this pattern dotted around the codebase, know that there's no reason for it, it's just down to me writing bad code.

iamlogand:
I'd prefer to merge these notifications:

_Senators in Faction 2 voted yea with 30 votes._

_Claudius added 22 votes to his own total from his influence as the Consul for Life nominee._

...so they look like this:

_Senators in Faction 2 voted yea with 30 votes. Claudius added 22 votes to his own total from his influence as the Consul for Life nominee._

But it's something that can be fixed separately. I'll raise an issue for it.

backend/rorapp/actions/attempt_assassination.py
```diff
@@ -154,8 +155,19 @@ def execute(
...
+        # Once elected or appointed, the Consul for Life cannot be assassinated (1.09.821)
+        if target.has_title(Senator.Title.CONSUL_FOR_LIFE):
+            return ExecutionResult(
+                False, "The Consul for Life cannot be assassinated."
+            )
+
+        # Only aligned senators are offered as targets (1.09.7), but execute is
+        # reachable with any senator id, so re-check before the faction lookup
+        # Only aligned senators are offered as targets (1.09.7), but execute is
+        # reachable with any senator id, so re-check before the faction lookup
```
iamlogand, line 167: nitpick: duplicated comment

iamlogand (approved):
Awesome!

backend/rorapp/actions/attempt_assassination.py
iamlogand: Fixed (directly on main)

PR #788, opened 2026-08-31
fix: recompute the HRAO when several senators die at once (§1.09.11)

### LINKED ISSUE #785: Epidemic kills senators in the wrong order, causing stale HRAO state
When an epidemic kills multiple senators including the HRAO, the victims list is fetched upfront and iterated without refreshing each object from the database. If the HRAO dies first and the title is passed to another victim mid-loop, that victim's in-memory object is stale and `kill_senator` never calls `set_hrao` after their death, leaving no one as HRAO. Two fixes are needed:

- Kill victims from lowest to highest rank so the HRAO is always last.
- Refresh each senator object from the database at the start of each loop iteration.

I've not tested yet, but I suspect the same bug will affect other situations that can kill multiple senators in one go, such as mob violence brought on by a badly recieved State of the Republic speech.

## Screenshot

Image

ActualRory:
Fixes #785.

When one event killed several senators, the victims were fetched upfront and iterated without reloading. If the HRAO died first, the title passed to another victim in the database, but the loop's stale copy of that victim still had no title, so kill_senator() never recomputed the HRAO after their death and Rome was left without one. The next forum phase then hung, because the first initiative goes to the faction holding the HRAO.

kill_senator() now reloads the senator, which fixes the succession at every call site. The new kill_senators() helper additionally kills a batch from lowest to highest rank, so the title passes once to a senator who survives rather than hopping across victims as they die. This isn't load-bearing, but keeps the logs from being full of HRAO succession.


iamlogand (approved):
Thanks for _sorting_ this issue!

PR #806, opened 2026-09-02
feat: add dev-only player buttons for local multiplayer

ActualRory:
Adds a dev-only button per player, to make testing easier. I've been using a form of these locally since I started working ofnt he project, but figure I may as well share for other folk to make use of. No idea what the rest of you have been doing!

### Behaviour

- One button per player in the game bar, naming the faction and the account holding it, with the current one marked. Clicking one signs the browser in as that player.
- Off by default, turned on from a new **Player Buttons** section in the Debug Tools panel. The choice is remembered in `localStorage`, because switching reloads the page and the buttons would otherwise disappear on every switch.
- Always shown in the NavBar. There is no debug panel outside a live game, so gating them there makes things fiddlier.

### Implementation notes

Hidden the same way the debug panel already is: the component probes `GET /api/test/presets/` on mount and renders nothing unless the backend answers. That route is registered solely under `TEST_ENDPOINTS_ENABLED`, so in production it does not exist and no control appears, the backend flag stays the single source of truth rather than a separate frontend build check.

Switching re-runs the test login and reloads. The session cookie is httpOnly and shared across the browser profile, so identity cannot be swapped in place; the reload is what rebuilds the game state and the websocket as the new player.

### Testing

Frontend only, no backend changes. From a logged-out browser, signed in via the buttons, created a game, joined all three factions by switching between them, started it, and switched faction in the live game bar; confirmed the buttons stay hidden until toggled and survive the reload.

player-buttons-debug-panel


iamlogand:
If you're wondering what I do: I usually test the game locally by having 3 chrome windows/profiles open, one linked to my own account and two linked to Google accounts created specifically for testing purposes. I just switch window to switch players.

ActualRory:
Interesting! I'm much lazier than that, as seen by this PR

iamlogand (approved):
Thanks for adding this. I'll use it when I can't be bothered loading up three Chrome instances

PR #811, opened 2026-09-04
Simplify pressure knight action description

### LINKED ISSUE #678: Remove dice reference from pressure knights action description
_Originally posted by @iamlogand in https://github.com/iamlogand/republic-of-rome-online/pull/665#discussion_r3411459494_

SoyMakari:
Closes #678

Shortens the Pressure knight action description and clarifies that each
pressured knight is removed after adding talents to its controlling senator's
personal treasury.

This is a text-only change. It was prepared in the cloud environment and was
not tested in a local clone; the repository's GitHub Actions checks will provide
the full validation.


ActualRory (approved):
Great change! Way more readable.

frontend/components/ActionDescription.tsx
```diff
@@ -31,11 +31,10 @@ const ActionDescription = ({ actionName, context }: ActionDescriptionProps) => {
   if (actionName === "Pressure knight") {
     return (
       <p>
-        Instead of attracting a knight, a player may pressure any knights they
-        control under their senators. Roll one die for each pressured knight and
-        add the total to the controlling senator's Personal Treasury. The
-        pressured knights are returned to the Bank and no longer provide extra
-        votes or income.
+        Instead of attempting to attract a knight, you may pressure your
+        senators' knights. Each pressured knight adds talents to its controlling
```
ActualRory, line 35: Actually, one thing, perhaps it should be 'each pressured knights adds 1d6 talents' or some other mention of the amount generated?

ActualRory (changes requested):

frontend/components/ActionDescription.tsx
ActualRory: Feel free to correct me if that value is shown elsewhere!

SoyMakari: Hi! If I’m not wrong the amount is shown immediately below as “up to X talents”, dynamically adjusted for Evil Omens. The review that led to #678 also specifically requested removing references to dice from user-visible text, so I think the current wording avoids duplication and keeps the description concise. But happy to include something like “ up to 6 random amount” or something like that if you think it would be better

ActualRory: Ah! You're absolutely correct! My apologies.

01-dialog
02-dialog-omens

ActualRory (approved):

Issue #775, opened 2026-08-29
Hover-Over Tooltips


ActualRory:
Many aspects of the game should be explained by hovering over them. This is a real dragon to slay and has been discussed before, but wanted to put it in issue form to reference later and remind me it's a thing worth doing.

Let's take logs - which is the first place that should have hover-over details about their contents.
For example, if a senator dies in the log - hovering over them should show their statline at the time of death, this way you can confirm who the hell just died and whether that matters!

More overarchingly, this would be one way to explain rules in motion. Though it's easy to overdo that.

Inspiration from crusader kings 3. 

<img width="674" height="722" alt="Image" src="https://github.com/user-attachments/assets/3d51eba3-6523-449e-8ac9-70e6b126bb89" />

<img width="654" height="891" alt="Image" src="https://github.com/user-attachments/assets/3ef6f836-7e7e-4644-85bf-3872a262a5ac" />

ActualRory:
Probably some overlap here with #753

ActualRory:
<img width="1280" height="720" alt="Image" src="https://github.com/user-attachments/assets/a6d2d7bf-4d0e-4536-8045-d43656c4b53d" />

iamlogand:
Big fan of the Crusader Kings II UI ❤️

ActualRory:
That one's crusader kings III! You casual 😁

iamlogand:
I just mistyped - I swear! 😆

PR #815, opened 2026-09-05
feat: hold a land victor in the field until the Revolution Phase (§1.11.3) (Rebel PR #1)

ActualRory:
Phase one of Rebel implementation. Moves the return of a victorious commander from the Combat Phase to the Revolution Phase, where §1.11.3 puts it. This is the groundwork for revolts: today a land victory dissolves the army inside `resolve_combat` before any declaration could be made.

> Starting with the HRAO and proceeding clockwise, each Commander who gained a Land Victory in this turn (including a Senate Victory) must either lay down his command (i.e., return his Force to the Active Forces boxes and return to Rome) or declare himself in Revolt. - §1.11.3

> A Land "Victory" lowers the Unrest Level by 1 and eliminates the War. Any other Commander on the War besides the victorious Commander returns to Rome and returns his Force to the Active Forces boxes. - §1.10.4

### Behaviour

- A land victor keeps his commander, legions and fleets through the end of the Combat Phase, and lays down his command in the Revolution Phase.
- Every other commander on that war still returns immediately, as §1.10.4 says.
- The victor is no longer made a proconsul. §1.10.8 grants that title only to a commander who "survives a non-victorious battle".
- His campaign cannot be recalled, reinforced or replaced while it waits.

Only the first of §1.11.3's two options is implemented here, which is why the new sub-phase is already named for the rule. Declaring a revolt follows in the next PR.

### New log examples

```
Scipio returned to Rome. Legions I, II and III returned to the reserve forces.
```

The Combat Phase says nothing extra about the victor staying: his army is visible in the UI, and this line already announces the return.

### Implementation notes

`Campaign.war` becomes nullable with `on_delete=SET_NULL` so the victor's campaign survives `war.delete()`, and `Campaign.land_victory` marks it. Campaigns with no war are skipped by the unprosecuted-war scan and by the recall, reinforce and replace proposals.

### Out of scope

Declaring a revolt, and everything that follows from it.

### Testing

585 backend tests pass, up from 567 on `main`. mypy is clean across 428 files.

image


docs/rule-interpretations.md
```diff
@@ -37,4 +37,16 @@ The implementation makes two decisions here:
 
 ## Combat phase
 
+### A land victory needs the enemy fleets beaten first (rule 1.10.4)
+
+"A Land 'Victory' [...] eliminates the War" (1.10.4) but a war that also shows a fleet strength requires two battles, and "the enemy Fleet Strength must be defeated before his land Forces may be attacked" (1.10.12). The implementation reads a land victory as ending the war only once the naval side has already been beaten, so a commander who wins the land battle of a war whose fleets are still at sea has not gained a Land Victory.
+
+This decides who may declare himself in revolt, because only "a Commander who gained a Land Victory in this turn" is offered that choice (1.11.3).
+
```
iamlogand: question: Can a commander win a land battle if there are still fleets at sea?

If not, then I'm not sure we need to document this as an interpretation, it's just a clear cut rule.

docs/rule-interpretations.md
```diff
@@ -37,4 +37,16 @@ The implementation makes two decisions here:
...
 ## Revolution phase
+
+### A land victor declares after cards are played (rule 1.11.3)
+
+The Declaration of Civil War is the third step of the Revolution Phase, after Play Statesmen/Concessions (1.11.1) and Excess Faction Cards (1.11.2), so the implementation holds the victorious commander in the field through card trading and keeps his army with him until he decides.
+
+Until then he is neither a Proconsul, since only "a Commander who survives a non-victorious battle" becomes one (1.10.8), nor subject to any senate proposal, since the Senate Phase has already passed. His Master of Horse, if he has one, stays with him rather than returning to Rome as he would if the Dictator had become a Proconsul (1.10.8), because the Master of Horse may join the revolt (1.11.32) and may fund the loyalty rolls (1.11.31).
```
iamlogand: suggestion: This covers facts that the rules aren't particularly ambiguous about.

I want to avoid filling this file with information that can already be found in the living rules without needing to make a decision one way or another.

iamlogand (commented):
Just looked at your new rule interpretations so far

docs/rule-interpretations.md
RoryAtSayari: True, actually! Will fix.

docs/rule-interpretations.md
ActualRory: Mea culpa!

backend/rorapp/models/campaign.py
iamlogand, line 1: idea: Instead of deleting wars once they are defeated, we could persist them, and when they are defeated we could set their status accordingly (via a new DEFEATED war status). I think this is a good time to make that change because for the first time we're interested in knowing when a campaign's war is defeated after the fact. What do you think of that?

A side note... For some time I've been thinking of making a similar change for senator death, where senators are never deleted after they die, allowing users to see all stats for dead senators. (totally informational, please don't feel the need to implement the senator change tomorrow 😆)

backend/rorapp/helpers/lay_down_command.py
iamlogand, line 1: praise: nice clean helper, I like it

backend/rorapp/effects/revolution_phase_end.py
```diff
@@ -8,14 +8,10 @@
...
+        return (
             game_state.game.phase == Game.Phase.REVOLUTION
-            and game_state.game.sub_phase == Game.SubPhase.PLAY_STATESMEN_CONCESSIONS
-        ):
-            return all(
-                f.has_status_item(FactionStatusItem.DONE) for f in game_state.factions
-            )
-        return False
+            and game_state.game.sub_phase == Game.SubPhase.CIVIL_WAR_DECLARATION
+        )
```
iamlogand, line 15: suggestion: Add a check here to ensure there are no pending commanders yet to lay down command.

What you have got currently works, but it subtly depends on the order of effects in the registry (`PlayCardsDoneEffect` being before `RevolutionPhaseEndEffect`), which we should avoid when we easily can.

iamlogand (changes requested):
Thanks for adding this. I know the idea is probably annoying, but let me know what you think!

I like the PR description. Not too long, just the right amount of detail.

Also liking the inclusion of semi-relevant artwork in the description 👍

iamlogand:
speech-bubble

iamlogand:
speech-bubble (2)

iamlogand:
speech-bubble (3)

ActualRory:
Absolutely creasing at your complimentary shitpost 🤣

ActualRory:
"Our beloved codebase is in the hands of mad men!"

backend/rorapp/models/campaign.py
ActualRory: This is a fair shout! I'd personally, long term, like to be able to look back through the logs - hovering over things to view the history of them in earnest.

image

I've shown that sort of thing before and outline it in #775 - but I think UI elements should provide context (historical or current) when hovered over. So if you hover over the mention of a defeated war, it should tell you in depth about it.

iamlogand: 💯

iamlogand: Yeah, preventing deletion of objects like wars, senators, etc would make these kinds of UI improvements possible in the future

ActualRory: Implemented. Wars are kept and marked DEFEATED now. Defeated wars are held back from the game state for now so nothing in the UI changes (yet!)

iamlogand (approved):

PR #816, opened 2026-09-05
feat: add the Declaration of Civil War and the civil war battle (§1.11.3, §1.11.31, §1.11.36 -> §1.11.373) (Rebel PR #2)

ActualRory:
**Phase two of rebel implementation.** ~~Depends on #815, so it must merge after it.~~
Migration `0063_rebels` depends on `0062_campaign_land_victory`. Until #815 merges this compare also shows its commit.

Gives a land victor the second of §1.11.3's two options and lets the Senate settle the matter in battle.

> Only one Faction may be in Revolt at a time. Should a Faction that wishes to Revolt field a stronger Army (inclusive of the respective Commanders' Military ratings) than that of a previously declared Rebel, the initial Rebel's declaration is ignored [...] The Army of the Primary Rebel or Governor is an Active War until the Rebel dies or achieves one of his Winning Conditions. - §1.11.3

> The Senate is the attacker and the Rebel the defender. The normal combat procedure is followed. The Strength number of the defender is the Strength of the Legions in his Army plus the Military rating of the Primary Rebel (which can never be higher than the Strength of his Army). - §1.11.37

### Behaviour

- Land victors decide in turn, starting with the HRAO's faction. Order ignores where senators actually are, because §1.11.3 says to assume every senator in play is in Rome.
- Before deciding, a commander may roll for his legions: one die each, following on a 5 or 6 in the Early Republic, with a talent buying +1 on a single legion. Veterans owing him allegiance follow without rolling; veterans owing allegiance to anyone else roll like the rest. Legions that refuse return to the reserve.
- A declaration creates the Civil War, moves the rebel to Italia and sends his fleets back to the reserve, since fleets play no part in a Civil War.
- A second faction is offered the declaration only when its army is strictly stronger. The earlier declaration is then ignored and that commander lays down his command instead. Once the phase ends the door closes: §1.11.3 lets nobody else revolt until the Primary Rebel is dead.
- The Senate deploys against the Civil War as against any other war. Losses fall on both armies, except that the rebel takes none in a Senate Defeat; mortality chits are drawn for the Senate's losses alone but can kill either side's commander.
- A Senate Victory kills every rebel senator and pays the commander half the rebel army's strength in popularity and influence. A stalemate carries the revolt into the next turn. The revolt also fails if the Primary Rebel dies by any route or loses his last legion.

### New log examples

```
Scipio spent 2T on the loyalty of Legion II and Legion III.
1 legion (II) refused to follow Scipio and returned to the reserve forces.
Scipio declared himself in revolt and is marching on Rome with 2 legions (I and III).
Scipio fielded the stronger army, so Fabius' declaration was ignored.
The revolt of Scipio was crushed.
```

### Implementation notes

The Civil War is an ordinary `War` row with `primary_rebel` set, so the 20T state debit, the four-active-wars losing condition, dictator eligibility, unprosecuted-war unrest and the deploy proposal all work without new code. Its land strength is the rebel army's own.

The combat results table and its loss rows move into `helpers/combat_results.py` so the ordinary and civil war resolvers share one table rather than keeping two copies.

### Out of scope

Secondary rebels, rebel markers and maintenance; the rebel's winning conditions. Rebel Governors are Advanced Rule §2.03 and need §2.02 first.

### Testing

628 backend tests pass, up from 585 on the branch below. `mypy .` is clean across 440 files.


ActualRory:
<details>
<summary>Obligatory art I like</summary> 

ghtwjp5pmpfg1_50

</details>

iamlogand (commented):
Ignore this comment - it was created by Claude via GH CLI and I can't delete it :/

backend/rorapp/models/war.py
```diff
@@ -27,6 +28,14 @@ class Status(models.TextChoices):
...
     status = models.CharField(max_length=12, choices=Status.choices)
     unprosecuted = models.BooleanField(default=False)
+    # A War with a Primary Rebel is the Civil War of his revolt (1.11.3)
+    primary_rebel = models.ForeignKey(
+        Senator,
+        related_name="revolts",
+        blank=True,
+        null=True,
+        on_delete=models.SET_NULL,
+    )
```
iamlogand, line 38: praise: good choice to use the existing table

backend/rorapp/actions/propose_replacing_proconsul.py
```diff
@@ -47,6 +47,7 @@ def get_schema(
                     c
                     for c in snapshot.campaigns
                     if not c.land_victory
+                and not c.rebel_army
```
iamlogand, line 50: nitpick: indentation is weird here and in propose_reinforcing_proconsul.py

backend/rorapp/helpers/civil_war.py
```diff
@@ -0,0 +1,194 @@
...
+        for c in game_state.campaigns
+        if c.land_victory and c.commander and c.commander.faction_id in positions
+    ]
+    if not campaigns:
+        return []
+
+    # Declaration order assumes that every senator in play is in Rome (1.11.3)
+    starting_senator = highest_ranking_senator(
+        [s for s in game_state.senators if s.faction_id and s.alive]
+    )
```
iamlogand, line 61: suggestion: rather than calling `highest_ranking_senator`, you should find the senator with the HRAO title.

I think `highest_ranking_senator` is only used when setting HRAO or changing presiding magistrate due to unanimous defeat -- moments where we can't rely on the HRAO title.

backend/rorapp/actions/declare_civil_war.py
```diff
@@ -0,0 +1,60 @@
...
+
+class DeclareCivilWarAction(ActionBase):
+    NAME = "Declare civil war"
+    POSITION = 0
+
+    def is_allowed(
+        self, game_state: GameStateLive | GameStateSnapshot, faction_id: int
+    ) -> Optional[Faction]:
+        campaign = declaring_campaign(game_state, faction_id)
+        if not campaign or not revolt_available(campaign):
```
iamlogand, line 24: issue (blocking): you're making three redundant DB calls via `revolt_available`:

- `revolt_available` > `standing_rebel` > `get_civil_war` > `War.objects`
- `revolt_available` > `standing_rebel` > `Senator.objects`
- `revolt_available` > `get_civil_war` > `War.objects`

iamlogand: issue (blocking): you're doing an additional DB call to `Senator.objects` via `revolt_available`:

- `revolt_available` > `standing_rebel` > `Senator.objects`

backend/rorapp/helpers/civil_war.py
iamlogand, line 1: suggestion: for readability and to avoid accidentally making DB calls within `action.is_allowed` or `effect.validate` implementations, you could move helpers that use in-memory game state to a separate file from those that make DB calls. In a future PR we can look at moving all in-memory helpers to a dedicated folder, but for now a separate file would be fine.

iamlogand (changes requested):
I made a light first pass where I just looked at some parts of the PR. I found some issues worth taking a look at.

Could you raise a new PR based on main with just the declare civil war and lay down command actions (excluding roll for legions and combat resolution changes)?

I'm working on the assumption that the split can be done without too much difficulty via Claude Code, since all the code is already written and in pretty good shape. Splitting out declaration would make this much easier to review, since I'm finding the size of this PR a bit overwhelming. Hope I'm not being unreasonable!

iamlogand:
@ActualRory I'm going to raise a PR to introduce some kind of feature flag mechanism so we can do trunk-based development of large features without breaking prod.

iamlogand:
@ActualRory - if you're up for splitting out declaration, I'd like to merge #850 first so you can put civil war declaration and delibrate laying down of arms behind the `civil_war` feature flag

ActualRory:
No worries at all on all of the above! I've split out declaration and laying down command into #852 (with the help of my friend Claude). It's all behind the civil_war flag now. 

Those inline comments are tackled there. Once #852 merges I'll rebuild this branch on top of it, with the rest behind the same flag. I've stuck a hold marker on this PR for now and I'll extract the rest of the code later, let's just do #852 for now.

iamlogand:
> No worries at all on all of the above! I've split out declaration and laying down command into #852 (with the help of my friend Claude). It's all behind the civil_war flag now. 
> 
> Those inline comments are tackled there. Once #852 merges I'll rebuild this branch on top of it, with the rest behind the same flag. I've stuck a hold marker on this PR for now and I'll extract the rest of the code later, let's just do #852 for now.

Sounds perfect! Thanks

PR #817, opened 2026-09-06
feat: add secondary rebels, rebel markers and maintenance (§1.11.32->§1.11.35, §1.06.2) (Rebel PR #3)

ActualRory:
**Phase three of rebel implementation.** Depends on #816, so it must merge after it: migrations `0064` and `0065` sit on top of `0063_rebels`. Until #815 and #816 merge, this compare also shows their commits.

Settles who else joins a revolt, what a revolt costs the rebels, and what their army costs to keep.

> After a Primary Rebel has been determined, each other Senator in his Faction must immediately declare his loyalty to the State or join him in Revolt. A Master of Horse may (not must) become a Secondary Rebel only if he is in the same Faction as a Rebel Dictator. (§1.11.32)

> Once the Primary and all Secondary Rebels are determined, each Senator in Revolt returns his Concessions to the Forum, loses all of his Knights and offices (except Governorships), and gains a Rebel marker. Rebels forfeit all future Personal Revenue. (§1.11.33)

> Rebel Legions & Fleets cost 2 Talents per turn to maintain and must be paid before the Redistribution of wealth occurs. (§1.11.35)

### Behaviour

- Every other living senator in the Primary Rebel's faction declares loyalty or joins him. A Master of Horse may join only alongside a rebel Dictator of his own faction, and the Consul for Life may never revolt (§1.09.823).
- A secondary rebel with an independent command leaves those forces to the Senate, since only the Primary Rebel's army marches on Rome.
- Each rebel returns his concessions, loses his knights and offices, and moves to Italia. He keeps his faction leader marker, which is not an office.
- Rebels generate no personal revenue, may neither give nor receive talents across the loyal/rebel line, and may not contribute to the State. Loyal senators of the faction may still pay into their own faction treasury.
- Veteran legions owing allegiance to any rebel desert to the Primary Rebel from wherever they are.
- Maintenance is 2 talents per rebel legion, paid before redistribution from the Primary Rebel's treasury, his faction's, or a secondary rebel's. Veterans owing allegiance to a rebel are free. Legions he cannot pay for are released to the Senate, and the HRAO decides whether the State takes them on; refused or unaffordable legions are eliminated.
- The State no longer pays maintenance for legions in rebellion.

### New log examples

```
Cornelius joined Scipio in revolt.
Valerius refused to follow the revolt and returned to Rome.
The revolt of Cornelius has made the mining concession available.
1 legion (IV) deserted to Scipio.
Scipio paid 6T to maintain 3 rebel legions.
[...]

PR #818, opened 2026-09-06
feat: add the rebel's winning conditions and the Rebel End Game (§1.12.1->§1.12.3) (Rebel PR #4)

ActualRory:
**Phase four of rebel implementation.** Depends on phase three (#817), so it must merge last: migration `0066` sits on top of `0065`. Until #815, #816 and #817 merge, this compare also shows their commits.

Completes the revolt: a rebel can now win, and Rome's own collapse becomes his victory rather than everyone's defeat.

> A result of "People Revolt" is obtained during a Population Phase with no Senator in Revolt. [...] The State Treasury goes Bankrupt with no Senator in Revolt. (§1.12.1)

> A Senator in Revolt, while Marching on Rome, defeats the Senate in battle without dying in that battle, or the Senate fails to attack his Army in a Combat Phase. Proceed to the Rebel End Game. (§1.12.2)

> If there are fewer than 4 Active Wars immediately after a Rebel achieves either of his Winning Conditions, he wins. If instead there are 4 or more Active Wars, skip directly to the Combat Phase [...] He then attacks Active Wars one at a time, in the order of his choosing. (§1.12.3)

### Behaviour

- Bankruptcy and a People Revolt end the game for everyone only when nobody is in revolt. With a rebel standing, he wins instead.
- A Senate Defeat the rebel survives wins outright, as does a Combat Phase in which the Senate never attacks him.
- The Era Ends tie-break counts influence from non-rebel senators only.
- Winning with four or more active wars sends him to the Rebel End Game: every surviving legion returns to the reserve, he takes them all along with every remaining fleet, and he attacks the active wars one at a time in an order he chooses, needing fleet support and a victory each time.
- Bringing Rome under four wars wins him the game even if he dies in the last battle, unless he won by beating the Senate. In that case dying there hands the game to a living Consul for Life, or the revolt fails, his Secondary Rebels are killed, and play continues. Failing to get under four, everyone loses.

### New log examples

```
The Senate sent no army against Scipio, so his march on Rome went unopposed.
Game over! Scipio seized Rome, so Faction 2 wins.
Game over! Scipio drove Rome's wars back below four, so Faction 2 wins.
Game over! Scipio fell in the last battle, so Faction 1 wins with its Consul for Life.
Game over! Scipio could not drive Rome's wars back below four, and the Republic collapsed.
Game over! The State Treasury fell into bankruptcy while Rome was in revolt.
```

### Implementation notes

The new game-over effects sit at the top of the effect registry, ahead of the existing bankruptcy check, because §1.12 orders a rebel's win above the losing condition it arises from. `Game.rebel_winning_condition` records which condition he met, since §1.12.3's last exception turns on it.

The four-active-wars losing condition already counted the Civil War, since it is an ordinary war row; it now discounts the war of a rebel who has already won.

### Out of scope

Rebel Governors (§2.03), which need Provincial Wars (§2.02) first, and captives (§1.10.71). The clauses in §1.11.32 and §1.12.2 that mention captives are no-ops until Capture exists.

### Testing

730 backend tests pass. mypy . is clean across 479 files.n the branch below. `mypy .` is clean across 456 files.


ActualRory:
Closing for the same reason as #817.

PR #820, opened 2026-09-06
feat: add both desertion events (§1.07.21)

ActualRory:
Adds both desertion events from the random events table (§1.07.21) and their dark blue sides: Ally Deserts / Roman Auxiliary Deserts, and Enemy's Ally Deserts / Enemy Mercenaries Desert.

> **Ally Deserts** - Roman allies are wavering. All battles fought this turn with an even result on 3d6 will result in a temporary increase to the War card's strength for this turn equal to the result on the black die. This increase is applied after any multipliers for Matching Wars.
>
> **Enemy's Ally Deserts** - All battles fought this turn with an odd result on 3d6 will result in a temporary decrease to the War's Strength for this turn equal to the result on the black die. This decrease is applied after any multipliers for Matching Wars. The minimum Strength it can be lowered to is 0.
>
> The dark blue sides read the white dice in place of the black die.

### Behaviour

- Rolling a 5 or a 16 on the random events table puts the matching desertion in play. A second draw flips the card to its dark blue side; a third does nothing, per §1.07.21.
- A battle's unmodified 3d6 decides which card speaks: an even roll strengthens the war, an odd roll weakens it. Both cards can be in play at once, but a roll is even or odd, so they never both fire on one battle and there is no stacking rule to model.
- The shift lands after the matching wars multiplier, alongside the leader's strength, so a doubled war is not doubled again.
- Disaster and standoff numbers are untouched, being read off the unmodified roll, as is the glory a victorious commander earns, which §1.10.4 fixes to the printed strength.
- Naval battles are affected as readily as land ones, against whichever strength is being fought.
- Both effects expire with the other temporary event effects at the start of the next forum phase. Since the forum phase precedes the combat phase within a turn, that is exactly the cards' "for this turn".

### Rule interpretation

Enemy's Ally Deserts says "the minimum Strength it can be lowered to is 0", which raises a question Ally Deserts never has to answer: does an enemy leader count toward that floor?

A leader raises the Land and Fleet Strength of his wars after any multiplier for matching wars (§1.07.342), and the desertion lowers them at that same point, so this implementation floors the war and its leaders as one strength. The Gladiator Revolt has a Land Strength of 6 and fights at 11 while Spartacus is in play; mercenaries deserting for 11 lower it to 0, not to 5.

The alternative reading would floor the printed strength alone and add the leader back afterwards, leaving his value untouched. Nothing in the rules singles the leader out here, unlike the Disaster and Standoff numbers, which §1.07.343 does keep separate. `docs/rule-interpretations.md` records the decision, and `test_enemy_desertion_cannot_lower_war_strength_below_zero` fails under the reading it rules out.

### New log examples

Drawing each event, then drawing it again:

> Faction 1 drew allied desertion. Rome's allies are wavering, so an even roll in any battle this turn will strengthen the enemy.

> Faction 1 drew allied desertion. With Rome's allies already wavering, Roman troops are shaken too, strengthening the enemy further on an even roll.

> Faction 1 drew enemy desertion. The enemy's allies are wavering, so an odd roll in any battle this turn will weaken the enemy.

> Faction 1 drew enemy desertion. With the enemy's allies already wavering, his mercenaries are deserting too, weakening him further on an odd roll.

Battles fought while each is in play, light blue side then dark blue:

> Rome's wavering allies deserted, strengthening the 1st Gallic War by 4.
> In a land battle, Cornelius' Campaign met with a stalemate, allowing the 1st Gallic War to continue. Legion 1, Legion 2 and Legion 3 were lost.

> Rome's shaken troops deserted, strengthening the 1st Gallic War by 4.

> The enemy's wavering allies deserted, weakening the 1st Gallic War by 3.
> In a land battle, Cornelius' Campaign won a victory, bringing an end to the 1st Gallic War. No legions were lost.

> The enemy's mercenaries deserted, weakening the 1st Gallic War by 9.

That last figure is the strength actually taken off, not what the dice offered: the war was fighting at 9 and the mercenaries deserted for 10, so the floor cut the desertion short and the log says 9.

And at the start of the next forum phase, from the existing expiry mechanism:

> The allied desertion has ended.

### Implementation notes

These are the first rules the engine has met that read one die out of a roll rather than the total, so `RandomResolver` grows `roll_dice_values`, returning the dice separately with the black die first. That is now the primitive both resolvers implement, and `roll_dice` sums it, so every existing caller is untouched.

`FakeRandomResolver` takes a list of dice where a single die's value matters, and keeps taking bare totals everywhere else: a queued total is spread across the dice it stands in for, preserving the sum. All 176 existing queue sites still queue the number they did before, including the deliberately impossible ones such as a 1 on 3d6. Only one annotation had to widen to match, in a test helper that forwards its argument to the queue.

`_get_desertion_strength` serves both cards through a parity argument rather than gaining a near-duplicate twin.

The `land_campaign` and `naval_campaign` fixtures move up to `tests/s1_basic_game/conftest.py`, since these tests fight a battle from the forum phase section and the testing conventions promote a fixture once a second section uses it. The move is verbatim; `two_campaigns` stays where it was.

No migration: `Game.effects` is a JSON field and `GameEffect` is a Python enum, so the new members need no schema change.

### Out of scope

**The combat calculator's odds do not account for either card.** `CombatCalculatorItem` reduces a battle to one flat modifier and reads the odds off a table keyed on the total of 3d6, so it cannot express a modifier that depends on which die showed what. While either card is in play its percentages are wrong, optimistic for allied desertion and pessimistic for enemy desertion. Fixing it means enumerating the 216 ordered outcomes rather than adjusting the number that goes in. I can raise this separately.

The war strength shown on the war card is correct as it stands and is deliberately left alone. It shows the strength a war fights at absent the dice, and there is no honest number to add before the roll. The effects panel carries the mechanic instead, as "+1d6 war strength on an even battle roll".

### Testing

`pytest` - 583 passed, up from 567. The sixteen new tests cover, for each card: the draw, the flip to the dark blue side, the third draw doing nothing, the black die, the white dice, and the roll of the wrong parity changing nothing. Plus the expiry, a naval battle, the strength floor, and a control battle with the same dice and no effect in play, which wins where allied desertion only draws.

mypy clean across 423 source files. `tsc --noEmit` on the frontend, clean.

<details>
<summary> Screenshots </summary>
image
image
image
</details>


ActualRory:
Improved the logs, they now look like this:

**Ally Deserts, drawn three times:**

> Faction 1 drew allied desertion. Rome's allies are wavering, so an even roll in any battle this turn will strengthen the enemy.

> Faction 2 drew allied desertion. With Rome's allies already wavering, Roman troops are shaken too, strengthening the enemy further on an even roll.

> Faction 1 drew allied desertion. Rome's allies are already wavering so there is no additional effect.

**Enemy's Ally Deserts, drawn three times:**

> Faction 2 drew enemy desertion. Enemy allies are wavering, so an odd roll in any battle this turn will weaken them.

> Faction 1 drew enemy desertion. With enemy allies already wavering, enemy mercenaries are deserting too, weakening them further on an odd roll.

> Faction 2 drew enemy desertion. Enemy allies are already wavering so there is no additional effect.


**Battle lines are as per below!**

> Rome's wavering allies deserted, strengthening the 1st Gallic War by 4.
> In a land battle, Cornelius' Campaign met with a stalemate, allowing the 1st Gallic War to continue. 3 legions (I–III) were lost.

> Enemy allies deserted, weakening the 1st Gallic War by 3.
> Enemy mercenaries deserted, weakening the 1st Gallic War by 9.
> In a land battle, Flaminius' Campaign won a victory, bringing an end to the Illyrian War. 2 legions (XI and XII) were lost. Unrest lowered by 1. The State Treasury gained 20T in spoils of war.

Still not exactly happy with these, will have another look before finishing the draft.

Open to suggestions!

backend/rorapp/classes/random_resolver.py
```diff
@@ -149,18 +155,24 @@ class FakeRandomResolver(RandomResolver):
     """
 
     def __init__(self) -> None:
-        self.dice_rolls: List[int] = []
+        self.dice_rolls: List[Union[int, List[int]]] = []
```
iamlogand, line 158: idea: With fake random resolver now supporting control on a per-die basis, the fake random resolver UI tool should also support this level of control, so that it can be used to manually exercise the desertion events.

Here's what it saves when I add "1, 2" (notice that the 2 is ignored):
Image

It should also have a little helper tooltop that explains which dice are black and white - not something I'd normally allow in the UI but acceptable for a debug tool.

I'll raise a separate issue for this since it's well out of scope for this PR.

backend/rorapp/helpers/handle_event.py
```diff
@@ -181,6 +185,58 @@ def handle_allied_enthusiasm(game: Game, current_faction: Faction) -> bool:
...
+
+    if level < 2:
+        game.add_effect(GameEffect.ALLIED_DESERTION)
+        game.save()
+
+    prefix = f"{current_faction.display_name} drew allied desertion."
+    if level == 0:
+        Log.create_object(
+            game.id,
+            f"{prefix} Rome's allies are wavering, so an even roll in any battle this turn will strengthen the enemy.",
```
iamlogand: issue: I'd prefer if you left out the bits that refer to dice rolls.

```suggestion
            f"{prefix} Rome's allies are wavering, so the enemy may be strengthened in battle.",
```

backend/rorapp/helpers/handle_event.py
```diff
@@ -181,6 +185,58 @@ def handle_allied_enthusiasm(game: Game, current_faction: Faction) -> bool:
...
+        )
+    elif level == 1:
+        Log.create_object(
+            game.id,
+            f"{prefix} With Rome's allies already wavering, Roman troops are shaken too, strengthening the enemy further on an even roll.",
```
iamlogand:
```suggestion
            f"{prefix} With Rome's allies already wavering, Roman troops are shaken too, so the enemy may be strengthened further in battle.",
```

backend/rorapp/helpers/handle_event.py
```diff
@@ -181,6 +185,58 @@ def handle_allied_enthusiasm(game: Game, current_faction: Faction) -> bool:
...
+
+    if level < 2:
+        game.add_effect(GameEffect.ENEMY_DESERTION)
+        game.save()
+
+    prefix = f"{current_faction.display_name} drew enemy desertion."
+    if level == 0:
+        Log.create_object(
+            game.id,
+            f"{prefix} Enemy allies are wavering, so an odd roll in any battle this turn will weaken them.",
```
iamlogand:
```suggestion
            f"{prefix} Enemy allies are wavering, so the enemy may be weakened in battle.",
```

backend/rorapp/helpers/handle_event.py
```diff
@@ -181,6 +185,58 @@ def handle_allied_enthusiasm(game: Game, current_faction: Faction) -> bool:
...
+        )
+    elif level == 1:
+        Log.create_object(
+            game.id,
+            f"{prefix} With enemy allies already wavering, enemy mercenaries are deserting too, weakening them further on an odd roll.",
```
iamlogand:
```suggestion
            f"{prefix} With the enemy's allies already wavering, their mercenaries are deserting too, so the enemy may be weakened further in battle.",
```

frontend/components/GameEffects.tsx
```diff
@@ -8,6 +8,12 @@ const EFFECT_FORMATTERS: Record<string, EffectFormatter> = {
     label: (level) => (level === 1 ? "Evil omens" : `Evil omens ×${level}`),
     annotation: (level) => "Harder military campaigns, easier persuasions",
   },
+  "allied desertion": {
+    label: (level) =>
+      level === 1 ? "Allied desertion" : "Roman auxiliary desertion",
+    annotation: (level) =>
+      `+${level === 1 ? "1d6" : "2d6"} war strength on an even battle roll`,
```
iamlogand: issue: this exposes too much of the mechanics (dice rolls) for my liking

```suggestion
      `War strength may be increased by up to {level === 1 ? "6" : "12"} in battle`,
```

frontend/components/GameEffects.tsx
```diff
@@ -32,6 +38,12 @@ const EFFECT_FORMATTERS: Record<string, EffectFormatter> = {
       level === 1 ? "Type III land bill" : `Type III land bill ×${level}`,
     annotation: (level) => `costs ${level * 10}T/turn`,
   },
+  "enemy desertion": {
+    label: (level) =>
+      level === 1 ? "Enemy desertion" : "Enemy mercenary desertion",
+    annotation: (level) =>
+      `−${level === 1 ? "1d6" : "2d6"} war strength on an odd battle roll`,
```
iamlogand:
```suggestion
      `War strength may be decreased by up to {level === 1 ? "6" : "12"} in battle`,
```

iamlogand (changes requested):
Good PR, but I've requested some user-visible text changes

backend/rorapp/helpers/resolve_combat.py
```diff
@@ -73,7 +83,29 @@ def resolve_combat(
...
-    evil_omens_level = Game.objects.get(id=game_id).count_effect(GameEffect.EVIL_OMENS)
+    game = Game.objects.get(id=game_id)
+    allied_level = game.count_effect(GameEffect.ALLIED_DESERTION)
+    allied_desertion = _get_desertion_strength(allied_level, dice, on_even=True)
+    if allied_desertion:
+        negative_modifier += allied_desertion
+        deserters = "wavering allies" if allied_level == 1 else "shaken troops"
+        Log.create_object(
+            game_id,
+            f"Rome's {deserters} deserted, strengthening the {war.name} by {allied_desertion}.",
```
iamlogand, line 95: praise: nice touch adding these logs, they look great!

iamlogand (commented):
Don't forget to check off README items

ActualRory:
Applied all of those! Added a $ because the suggestion as-written would render the braces literally.

iamlogand (dismissed):

iamlogand (approved):

PR #821, opened 2026-09-06
feat: add repopulating Rome (§1.09.81)

### LINKED ISSUE #732: Reappearing senator doesn't join statesman
A senator can reappear as an unaligned senator when there is a matching statesman already in play. This is incorrect - the senator should have joined the statesman instead.

ActualRory:
Adds Repopulating Rome (§1.09.81), which tops the senate back up when deaths thin it out during a senate phase, together with the statesman exception §1.05.4 states for the same stack of cards.

Edit: Closes #732 now too, because the same statesman comparison settles what happens when a family card is revived while putting Rome in order (§1.07.8).

> Whenever the number of Aligned Senators currently in Rome is below 8 during a Senate Phase, the top Senator card in the Curia is promoted to active status in the Faction of the player with the fewest total Senators. Ties are resolved in favor of the player with the least total Influence currently in Rome, with further ties being broken by a dice roll. Such promotions continue until the number of active Senators in Rome reaches 8 or the Curia is empty. If the Curia has no Senators, those in the Forum can be substituted. The choice of which of the available Senators to take is up to the receiving player. If no further Senators are currently available in either the Curia or the Forum, no such promotion occurs.

### Behaviour

- While a senate phase has fewer than 8 aligned senators in Rome, the faction with the fewest senators gains one. Ties are settled by the least influence in Rome, then by a dice roll.
- The senator who died first is promoted first, returning as the next generation of his family.
- When no dead senator is available, the receiving player takes any unaligned senator instead, through a new **Select senator** action that suspends the senate until it is answered.
- Promotions repeat until Rome holds 8 aligned senators or nothing is left to take.
- A dead senator whose statesman is already in play is passed over in favour of the next one down (§1.05.4), because a family card cannot be promoted away from its statesman (§1.07.312).
- A family card revived while putting Rome in order joins a statesman of its family rather than standing up as a second senator of the same house (§1.07.312, §1.07.8). The statesman takes the card and the heir's generation, so the count is still right when the statesman later dies and the family card is all that is left.

### New log examples

Fulvius died before Aelius, so Fulvius is the one promoted:

```
Julius of Faction 3 died in the epidemic. His heir Julius II replaced him as faction leader.
Fulvius of Faction 3 died in the epidemic.
Aelius of Faction 3 died in the epidemic.
With fewer than 8 aligned senators in Rome, Fulvius' heir Fulvius II joined Faction 3.
```

When nobody is waiting and the choice falls to the player:

```
Aelius' heir Aelius II became an unaligned senator.
With fewer than 8 aligned senators in Rome, Aelius II joined Faction 3.
```

Cornelius has a statesman in play and Fabius does not, so only Fabius stands alone:

```
The Latium tax farmer concession was restored and can be awarded again.
Cornelius' heir joined P. Cornelius Scipio Africanus.
Fabius' heir Fabius II became an unaligned senator.
```

### Implementation notes

- `RepopulateRomeEffect` needs no loop of its own. The effect executor re-runs every effect until none validates, so promoting one senator per pass gives the rule's "promotions continue until" for free.
- `Senator.curia_position` records death order, since the rule reads the top of a stack. It is stamped when a family card without a faction leader marker dies, and cleared whenever that senator returns to play, including through the existing revival while putting Rome in order.
- The check stands down once the senate closes, because the automatic recalls there (§1.09.9) may bring commanders home and cover the shortfall by themselves.
- The comparison that finds a statesman for a family code existed twice already, once for the deck draw and once for the promotion order, so it moves into `statesman_in_play` and all three callers share it. A statesman played while its family card is dead leaves a row with `family` false, which is what the three of them are looking for.
- The player's choice reuses the existing interrupt slot, so the senate resumes in the exact sub-phase it paused in. Assassination stays available throughout, as §1.09.7 allows.

### Out of scope

- The rest of §1.05.4: elimination itself, the immediate draw for a faction left with no senators, and the option to quit instead.
- Players cannot see who is next in line, since dead senators are listed alphabetically rather than in the order they died.

### Testing

`pytest` 581 passed, 13 of them new in `s1_09_81_repopulating_rome_test.py` and one in `s1_07_8_putting_rome_in_order_test.py`. `mypy` clean across 428 source files.

### Screenshots

shot2
shot3


backend/rorapp/actions/select_senator.py
```diff
@@ -0,0 +1,101 @@
...
+        game_id: int,
+        faction_id: int,
+        selection: Dict[str, Any],
+        random_resolver: RandomResolver,
+    ) -> ExecutionResult:
+        game = Game.objects.get(id=game_id)
+        faction = Faction.objects.get(id=faction_id)
+        senator = Senator.objects.get(id=selection["Senator"])
+
+        if not senator.alive or senator.faction_id:
```
iamlogand: issue: also ensure the senator is in Rome

backend/rorapp/effects/repopulate_rome.py
```diff
@@ -0,0 +1,77 @@
...
+        if game.phase != Game.Phase.SENATE:
+            return False
+        if game.sub_phase in (
+            Game.SubPhase.START,
+            Game.SubPhase.ASSASSINATION_RESOLUTION,
+            Game.SubPhase.REPOPULATION,
+            # The senate closes with automatic recalls, which may fill Rome (1.09.9)
+            Game.SubPhase.END,
+        ):
+            return False
```
iamlogand: praise: good idea - things would get messy without this!

backend/rorapp/effects/repopulate_rome.py
```diff
@@ -0,0 +1,77 @@
...
+
+class RepopulateRomeEffect(EffectBase):
+
+    def validate(self, game_state: GameStateSnapshot) -> bool:
+        game = game_state.game
+        if game.phase != Game.Phase.SENATE:
+            return False
+        if game.sub_phase in (
+            Game.SubPhase.START,
+            Game.SubPhase.ASSASSINATION_RESOLUTION,
```
iamlogand: issue: The `interrupted_sub_phase` is already used to store the paused subphase during the `SPECIAL_MAJOR_PROSECUTION` subphase, so this effect could fire during the special major prosecution and make a mess of things, probably breaking the game.

iamlogand (changes requested):
Looks like a solid approach, cheers! Happy to merge after these two small things are resolved

iamlogand (approved):

PR #822, opened 2026-09-06
feat: add the mob violence event (§1.07.21)

ActualRory:
Adds the mob violence event, the 3 result on the early republic random events table (§1.07.21). It was already named in `event.json` but had no handler, so drawing it fell through to the unimplemented-event path and the player drew a forum card instead.

> **Mob Violence**
>
> - Draw Mortality Chits equal to the current Unrest Level.
> - Only those Senators currently in Rome whose Popularity is less than the current Unrest Level can be killed.
>
> **Dark blue side: More Mob Violence**
>
> - Draw Mortality Chits equal to the current Unrest Level + 1d6.
> - Only those Senators currently in Rome whose Popularity is less than the current Unrest Level can be killed.
> - Further Mob Violence events are resolved as More Mob Violence events.

### Behaviour

- Drawing the event puts the card in play and draws mortality chits equal to the unrest level.
- A senator is killed only if his chit comes up, he is in Rome, and his popularity is below the unrest level. A senator as popular as the unrest level is safe however many chits are drawn.
- At unrest 0 the light blue side draws no chits at all, so nobody is at risk even though a senator with negative popularity would otherwise qualify.
- A second draw flips the card to more mob violence, which adds 1d6 to the chits drawn. Every later draw is resolved the same way, so the card never reverts.
- Deaths reuse the existing mob cause of death, so they read the same as the ones from a state of the republic speech and release concessions to the forum like any other death.

### New log examples

A first draw with the unrest level at 4, killing two senators:

> Faction 1 drew mob violence. A mob rioted in Rome, turning on senators less popular than the unrest level of 4.
> Claudius of Faction 2 was killed by the mob.
> Cornelius of Faction 1 was killed by the mob.

A second draw, where nobody's chit comes up:

> Faction 1 drew mob violence. With Rome already prone to mob violence, an even larger mob rioted, turning on senators less popular than the unrest level of 4, but every senator survived.

A draw while Rome is calm:

> Faction 1 drew mob violence. With the unrest level at 0, the mob dispersed without violence.

### Implementation notes

- `handle_mob_violence` sits with the other event handlers in `handle_event.py`, and `GameEffect.MOB_VIOLENCE` records whether the card has flipped, the same way natural disaster tracks its own escalation.
- The extra 1d6 on the dark blue side is reduced by 1 per evil omens card in play, down to a minimum of no extra chits. §1.07.21 puts a -1 on every die roll except initiative and events table rolls, and the only roll it goes on to exempt by name is the one that picks a concession for a natural disaster. This is recorded in `docs/rule-interpretations.md` because it has the odd consequence of making a riot less deadly, which cuts against the spirit of the card.
- The README's event checklist had drifted from the early republic table. Barbarian raids and pretender emerges are middle and late republic events that cannot be drawn in this scenario, and mob violence was missing from the list entirely, so the list now matches the table in `game-data/components/board.md`.

### Out of scope

- The rest of the unimplemented early republic events: refuge, storm at sea, new alliance, rhodian alliance, enemy leader dies, trial of verres, and the two desertion events already covered by #820.

### Testing

- 7 new tests in `s1_07_21_events_test.py`: the event fires on a 3, kills an unpopular senator in Rome, spares one as popular as the unrest level, spares one away from Rome, draws nothing at unrest 0, escalates to more mob violence on a second draw, draws its extra chits even with no unrest, and has those extra chits cancelled by three evil omens.
- `pytest`: 574 passed.
- `mypy`: no issues found in 422 source files.


SoyMakari:
Hi @ActualRory , in the Spanish version of this card the dark blue side kills senators with popularity_limit = unrest + 1 , is it the same in the English version?

ActualRory:
> Hi @ActualRory , in the Spanish version of this card the dark blue side kills senators with popularity_limit = unrest + 1 , is it the same in the English version?

Hey! Yes. There's a protection cap, where if you're more popular than the current unrest level - you're all good. (or the unrest level+1 for the dark blue side, as you say.) Still, their mortality chit has to be drawn.

ActualRory:
Oh, you're absolutely right in that this isn't reflected in the rules - seems like a mistake in transcribing the rules to the repo, which I carried forward on this one. Great catch!

backend/rorapp/helpers/handle_event.py
```diff
@@ -306,3 +308,52 @@ def handle_natural_disaster(
...
+    # risk, however many chits are drawn (1.07.21)
+    senators = Senator.objects.filter(
+        game=game.id, alive=True, location="Rome", popularity__lt=threshold
+    )
+    victims = [s for s in senators if get_senator_codes(s.code)[0] in codes]
+
+    prefix = f"{current_faction.display_name} drew mob violence."
+    if chit_count == 0:
+        message = (
+            f"{prefix} With the unrest level at 0, the mob dispersed without violence."
```
iamlogand: nitpick: We've never actually described it as "unrest level" in the UI AFAIK. Let's save this decision for the future and just call it "unrest" for now.

```suggestion
            f"{prefix} With unrest at 0, the mob dispersed without violence."
```

frontend/components/GameEffects.tsx
```diff
@@ -48,6 +48,9 @@ const EFFECT_FORMATTERS: Record<string, EffectFormatter> = {
     label: (level) => (level === 1 ? "Drought" : "Severe drought"),
     annotation: (level) => `famine severity +${level}`,
   },
+  "mob violence": {
+    label: (level) => (level === 1 ? "Mob violence" : "More mob violence"),
```
iamlogand: nitpick: While "more" is true to the source material, it's too comparative in this context.

```suggestion
    label: (level) => (level === 1 ? "Mob violence" : "Rampant mob violence"),
```

Open to other suggestions

backend/rorapp/helpers/handle_event.py
iamlogand: nitpick: We've never actually described it as "unrest level" in the UI AFAIK. Let's save this decision for the future and just call it "unrest" for now.

Also, this log isn't entirely accurate. It's possible for `chit_count` to be zero while unrest is greater than zero, due to evil omens. I'd suggest this:

```suggestion
            f"{prefix} Due to low unrest, the mob dispersed without violence."
```

backend/rorapp/helpers/handle_event.py
```diff
@@ -306,3 +308,52 @@ def handle_natural_disaster(
...
+        )
+    else:
+        if level == 0:
+            message = f"{prefix} A mob rioted in Rome, turning on senators less popular than the unrest level of {unrest}"
+        else:
+            message = f"{prefix} With Rome already prone to mob violence, an even larger mob rioted, turning on senators with a popularity below {threshold}"
+        message += "." if victims else ", but every senator survived."
```
iamlogand: idea: Instead of exposing the unrest and threshold numbers in the log, here's an alternative:

Do something similar to what `GiveSpeechAction` already does, and regardless of mob level, log "An outraged mob rioted in Rome, killing N senators."

If the popularity threshold was of no significance, leave it at that.

However, if any senators IDs were drawn, but they were spared due to their popularity, it would be fun to log that:
"Fabius was caught by the mob but quickly released thanks to his popularity."
"Fabius and Terentius were caught by the mob but quickly released thanks to their popularity."

This exposes the popularity threshold mechanic in the only way it matters, while saving players from having to lookup senator popularities themselves.

I appreciate this means tweaking the senator retrieval code higher up, but I think it'll be worth it.

iamlogand (changes requested):
Looks good, just some suggestions related to user visible text. Let me know what you think of my idea.

ActualRory:
Applied all your suggestions, thanks!

ActualRory:
Ah, just spotted the idea! Standby, will have a look at that later/tomorrow and you can review after I either action that one or push back

ActualRory:
Behold:
```
Faction 1 drew mob violence. An outraged mob rioted in Rome.
Fabius of Faction 3 was killed by the mob.
Cornelius of Faction 1 was killed by the mob.
Terentius of Faction 2 was caught by the mob but quickly released thanks to his popularity.
```

If noone gets caught:
`Faction 1 drew mob violence. An outraged mob rioted in Rome, but every senator survived.`

iamlogand (dismissed):
~Looks like my code suggestion for GameEffects.tsx had a syntax error. Will re-approve once that's fixed.~
Fixed and merged

iamlogand (approved):

PR #826, opened 2026-09-07
fix: correct counter-bribe recipient text

### LINKED ISSUE #825: Counter-bribe dialog incorrectly says talents are transferred to the persuader
The counter-bribe dialog says that Talents are transferred from the spending faction's treasury to the persuading senator.

According to Living Rules v1.07B §1.07.413, counter-bribes are paid into the Personal Treasury of the target senator. The backend already transfers the Talents to the correct senator, so only the explanatory text in the dialog needs to be corrected to refer to the target instead of the persuader.

Image

This appears to be only a wording error in the dialog; the Talents are already transferred correctly to the target Senator.

I can try to fix this.

SoyMakari:
Closes #825

The counter-bribe dialog incorrectly stated that Talents were transferred to the persuading Senator. It now correctly identifies the target Senator as the recipient.

The backend already transfers the Talents to the target correctly, so no game logic has been changed.

## Testing

- Prettier check
- `git diff --check`
- Frontend production build completed successfully in a separate environment

### DIFF
--- a/frontend/components/customActionForms/CounterBribeForm.tsx
+++ b/frontend/components/customActionForms/CounterBribeForm.tsx
@@ -87,7 +87,7 @@ const CounterBribeForm = ({
             <h3 className="text-xl">Counter-bribe</h3>
             <p>
               Prevent persuasion by transferring talents from your faction
-              treasury to {persuader?.displayName} as a counter-bribe.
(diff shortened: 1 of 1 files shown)


ActualRory (approved):
Great catch! Nice fix

PR #832, opened 2026-09-10
refactor: use the pluralize helper for count labels

### LINKED ISSUE #774: Use the frontend `pluralize` helper
There are many places where `pluralize()` should be used.

For example: `{senator.votes === 1 ? "vote" : "votes"}`

_Originally posted by @iamlogand in https://github.com/iamlogand/republic-of-rome-online/pull/769#discussion_r3887739249_

ActualRory:
Closes #774.

`pluralize(count, noun)` has lived in `frontend/helpers/text.ts` since the legion and fleet counters were added, but most of the frontend still does it by hand, so to speak. This replaces all(?) 18 remaining sites with the helper.

### Behaviour

No rendered text changes. Three sites tested `count > 1` rather than `count !== 1`, so they would have rendered `0 legion`, but all three sit inside a `count > 0` guard and the case was unreachable.

### Implementation notes

The helper suffixes the whole noun phrase, so `pluralize(n, "more regular legion")` gives `3 more regular legions` and no site needed a special case. Two read a little differently after the swap:

- `RedistributeTalentsForm` keeps its `Total: 3 / 5 talents` shape by leaving `allocTotal` outside the call: `Total: {allocTotal} / {pluralize(total, "talent")}`.
- `GameMain` and `AdvancedVoteForm` lose a `{" "}` separator each, since the helper emits the space between the count and the noun itself.

`helpers/date.ts` accounts for 6 of the 18 (`min`, `hour`, `day`, `week`, `month`, `year`); the rest are spread across `CombatCalculator`, `GameBar`, `GameMain`, `AdvancedVoteForm`, `PlaySecretBodyguardForm` and `RedistributeTalentsForm`.

`CombatCalculator.tsx` is not prettier clean on `main`, so it was edited by hand rather than run through prettier, which keeps the diff to the lines that actually changed.

### Out of scope

`GameEffects.tsx` has a similar looking cluster of `level === 1 ? ... : ...` ternaries, but those pick between different labels (`Drought` / `Severe drought`, `Evil omens` / `Evil omens x2`) rather than singular and plural, so the helper does not apply there.

### Testing

`npx tsc --noEmit` clean, `npm run build` succeeds, `npx eslint` clean on all 7 changed files. `npm run lint` is already broken on `main` (`next lint` was removed in Next 16), so eslint was invoked directly

### DIFF
--- a/frontend/components/CombatCalculator.tsx
+++ b/frontend/components/CombatCalculator.tsx
@@ -19,6 +19,7 @@ import {
   createProposalCalculation,
   getDeployedForces,
 } from "@/helpers/deploymentProposal"
+import { pluralize } from "@/helpers/text"
 import useIsMobile from "@/hooks/isMobile"
 
 import CombatCalculatorItem from "./CombatCalculatorItem"
@@ -363,8 +364,7 @@ const CombatCalculator = forwardRef<CombatCalculatorHandle, GenericActionFormPro
 
         if (additionalLegionsNeeded > 0) {
           unitsNeeded.push(
-            `${additionalLegionsNeeded} more regular legion` +
-              (additionalLegionsNeeded > 1 ? "s" : ""),
+            pluralize(additionalLegionsNeeded, "more regular legion"),
           )
         }
       }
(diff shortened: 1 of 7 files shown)


iamlogand (approved):
Appreciate you taking the time to clean things up, cheers

Issue #823, opened 2026-09-07
Persuasion rejects 0% attempts and Blackmail cannot reduce Popularity below 0


SoyMakari:
Hi,

While manually testing Persuasion Attempts with Blackmail, I found two related issues:

1. The game prevents a Persuasion Attempt when the calculated success chance is 0%.
2. When Blackmail fails, the target Senator's Popularity cannot fall below 0, although negative Popularity is allowed by the rules.

These interact in an important way because Blackmail can have strategic value, reducing an opposing Senator's Popularity and Influence, even when the Persuasion Attempt itself has a 0% chance of succeeding  (at least, this is the way I understand this card and how we play it, maybe it's just we misunderstood the rules, but I don't see anything against this in the rules)

### Persuasion Attempts — §1.07.4

Rule 1.07.4 defines a legal Persuasion Attempt by the status and location of the Senators:

- the persuading Senator must be in Rome;
- the target may be an Unaligned Senator in the Forum; or
- an already Aligned non-Faction Leader Senator who is also in Rome.

The rule does not state that a Persuasion Attempt requires a minimum positive probability of success.

### Resolution — §1.07.41

Rule 1.07.41 calculates a Base Number from the persuader's Oratory and Influence against the target's Loyalty, modified as applicable, and compares that Base Number with a 2d6 roll.

There is no additional requirement in §1.07.41 that the resulting Base Number must provide a greater-than-zero chance of success before the attempt may be made.

### Bidding — §1.07.414

Rule 1.07.414 says that the persuader announces the target and the initial Base Number and may eventually roll 2d6 at the current Base Number.

The example in this section explicitly begins with Fabius having a Base Number of 0 before spending Talents.

I therefore cannot find a rule in §§1.07.4–1.07.414 that makes an otherwise legal persuader/target combination illegal solely because its current success chance is 0%.

This becomes particularly relevant when Blackmail is available.

### Blackmail — §1.07.35 / Blackmail card text

Under §1.07.35, Intrigue cards are used according to their specific restrictions and effects.

Blackmail makes the current Persuasion Attempt unopposed and, if it fails:

> "the target Senator's Influence and Popularity are each reduced by 2d6."

Therefore a failed Persuasion Attempt can itself be strategically useful when Blackmail is played: even if the target is extremely difficult or impossible to persuade, failure can reduce that rival Senator's Influence and Popularity.

The current implementation correctly applies these losses to the **target Senator**. That part is correct and should remain unchanged.

[...]

PR #833, opened 2026-09-10
fix: persuasion odds and blackmail penalties (§1.07.4, §1.07.31)

ActualRory:
Fixes the three persuasion bugs reported in #823, plus one found while tracing them.

> Popularity may not go above 9 or below -9. (§1.07.31)

> For your Persuasion Attempt on your initiative, make an unopposed Persuasion Attempt against a non-Faction Leader Senator of your choice. [...] If it fails, the target Senator's Influence and Popularity are each reduced by 2d6. (Blackmail)

> The State Treasury must immediately pay 20T for sacrifices and Temple repair and a −1 penalty is applies to EVERY die or dice roll except for Initiative and further rolls on the Events table. (Exception: +1 to all Persuasion Attempts.) (Evil Omens)

### Behaviour

Blackmail can push Popularity below zero. A failed blackmailed attempt reduced the target's Popularity with a floor of 0 rather than the -9 of §1.07.31, so the reporter's target sat at 0 Popularity and lost nothing at all from a roll of 7. The floor is now -9. Influence keeps its floor of 0, which is correct.

A 0% attempt is legal when Blackmail is used. Nothing in §1.07.4 to §1.07.414 makes an otherwise legal persuader and target illegal because the attempt cannot succeed, and with Blackmail in hand a hopeless attempt is worth making for the 2d6 Influence and Popularity it costs the target. The backend gate, the frontend candidate filter and the disabled Confirm button all stand down when Blackmail is used. Without Blackmail the attempt is still refused, as Logan intended.

The persuasion step is no longer skipped past a possible Blackmail. The automatic skip fired whenever the best available attempt was at 0%, which took the whole step away from a player holding the card.

Evil Omens gives +1 to persuasion attempts, but only the actual resolution roll was counting that. The 0% gate, the success chances logged by Attempt persuasion, Counter-bribe and Continue persuasion, and the automatic skip all built the base number without the bonus. With omens in play the engine could refuse an attempt that had a chance, understate the odds shown to the players, and skip the step when a legal attempt was available.

### New log examples

The blackmail penalty line reports the loss actually applied rather than the roll, matching how sponsored games, land bills and combat glory already log theirs. Fulvius holds 1 Influence and -5 Popularity, and the rolls are 3 and 7:

> Cornelius failed to persuade Fulvius.
> Blackmail: Fulvius lost 1 influence and 4 popularity.

Before this change the same line read `lost 3 influence and 7 popularity`, and Fulvius stayed on 0 Popularity.

With one Evil Omens event in play, the chance quoted at the start of an attempt now matches the roll that resolves it:

> Cornelius of Faction 1 began a persuasion attempt targeting the unaligned senator Testius (3% success chance).

The same correction applies to the counter-bribe and continue-persuasion lines, which quote a chance in the same way.

### Implementation notes

The base number was written out five times, which is why one copy could carry the omens bonus and the other four could not. It now lives in `persuasion_modifier`, with the omens level as a required argument so a call site cannot quietly leave it out. `PersuasionAutoSkipEffect` finds the best pairing by asking the helper for each persuader and target rather than combining a best persuader score with a lowest target score, so it gets the same answer from the same formula as every other caller.

The popularity floor reuses `Senator.change_popularity`, which already clamps to the -9 to 9 range and returns the change actually applied.

The auto-skip cannot ask whether the faction holds Blackmail, because hands are secret (§1.04.9) and a skip that read the card would announce it to the table. Card counts are already public (#759), so it keys on the count instead: a faction holding any card is offered the persuasion step even when its best chance is 0%. A faction holding three irrelevant cards now has a pointless step to skip, which seemed the better trade than leaking the hand.

### Out of scope

The frontend already applied the omens bonus in both its candidate filter and its displayed odds, so no change needed there.

### Testing

609 backend tests pass, 7 of them new, and `mypy .` is clean over 433 files. The three omens tests were confirmed to fail with the bonus removed, so they hold the fix down. On the frontend, `tsc --noEmit` and prettier are clean.


backend/rorapp/effects/persuasion_auto_skip.py
iamlogand, line 1: idea: Now that autoskip can be blocked by blackmail, this effect can reveal to highly observant opponents when a card-holding faction has no blackmail card. This would be detrimental if the faction wished to lie about whether they have the card. Therefore I suggest we delete this effect entirely and make every player manually skip persuasion every time.

iamlogand (approved):
Good to merge! Going to raise a separate issue for the blackmail autoskip removal idea

backend/rorapp/effects/persuasion_auto_skip.py
ActualRory: I'd counterpoint this with the system from Magic The Gathering: Arena. I'll bring this up in the issue!

PR #835, opened 2026-09-10
feat: Add the storm at sea event (§1.07.21)

### LINKED ISSUE #834: Implement the Storm at Sea event for the Basic Game
## Summary

Implement the **Storm at Sea** event according to the multiplayer **Living Rules v1.07B**, using the Basic Game rules currently supported by the application.

The event should eliminate Roman Fleets according to a 2d6 roll. When a choice is required, the HRAO must select the fleets to be destroyed.

## Rules scope

The lack of an operational province system should not block the implementation of this event.

Provincial Fleets belong to the optional Advanced Game rules and are not used in the Basic Game. This implementation should therefore affect only standard Roman Fleets and intentionally leave Provincial Fleets out of scope.

## Expected behaviour

When Storm at Sea is drawn:

* Roll 2d6 to determine the number of Roman Fleets lost.
* Apply the existing Evil Omens modifier to the event roll.
* Count both fleets in the reserve and fleets assigned to campaigns.
* Never require more fleets to be destroyed than currently exist.
* If Rome has no fleets, record that there were no fleets to lose and continue the Forum Phase.
* If the number of losses is equal to or greater than the number of existing fleets, destroy all existing fleets automatically.
* Otherwise, require the HRAO to select exactly the necessary number of fleets.
* Only a player controlling the HRAO's faction may resolve the selection.
* Persist the destroyed fleets normally so that subsequent events and phases see the updated fleet state.
* Allow the event to occur and resolve correctly more than once during the same turn.
* Do not prevent Rome from raising new fleets later.

The event does not directly recall a campaign. If destroyed fleets leave a land campaign without sufficient Fleet Support:

* Record a warning explaining that the campaign no longer has sufficient Fleet Support.
* Allow the Senate to restore the necessary support by raising or assigning fleets.
* At the end of the Senate Phase, use the existing campaign validation procedure to recall a campaign that still lacks the required Fleet Support.

No separate recall marker or parallel campaign-support system should be introduced if the existing Senate procedure can make this determination from the current fleet assignments.

## HRAO interface

When manual selection is necessary, display the available fleets as checkboxes in this order:

1. Fleets in the reserve, ordered by fleet number.
2. Fleets assigned to campaigns, grouped by campaign and ordered by fleet number within each group.

The interface should:

* Show the number of fleets that must be selected.
* Prevent submission until exactly that number has been selected.
* Be available only to the faction controlling the HRAO.
* Make the Basic Game limitation regarding Provincial Fleets explicit.

If every remaining fleet must be destroyed, the result should be resolved automatically without presenting an unnecessary selection window.

## Game log

The game summary should record:

* That Storm at Sea was drawn.
* The resulting number of losses.
* That the HRAO must choose the fleets, when applicable.
* The exact Roman Fleets destroyed.
* That Rome had no fleets to lose, when applicable.
* Any campaign that has lost sufficient Fleet Support.
* Any later automatic recall caused by insufficient support, using the existing Senate log behaviour.

## Acceptance criteria

Backend and frontend tests should cover:

* Selection restricted to the HRAO's faction.
* Selection of the exact required number of fleets.
* Reserve fleets displayed before deployed fleets.
* Numerical ordering within each group.
* Destruction of both reserve and deployed fleets.
* Automatic destruction when all remaining fleets must be lost.
* Resolution when no Roman Fleets exist.
* Repeated Storm at Sea events.
* Correct persistence of fleet losses.
* Warning when a campaign loses sufficient Fleet Support.
* Reinforcement of an affected campaign during the Senate.
* Automatic recall at the end of the Senate when support is not restored.
* Explicit exclusion of optional Provincial Fleets.
* No regressions in the existing Forum, Senate and Combat Phase behaviour.

SoyMakari:
Closes #834.

## Summary

Implements the Early Republic **Storm at Sea** event for the Basic Game, following the multiplayer Living Rules v1.07B.

The event makes a separate 2d6 effect roll, applies Evil Omens and destroys the resulting number of existing Roman Fleets. When a choice is required, only the faction controlling the HRAO may select the fleets.

## Behaviour

* Rolls 2d6 for the number of Roman Fleets lost.
* Reduces the result by 1 for each Evil Omens effect in play, to a minimum of 0.
* Counts Roman Fleets both in the reserve and assigned to campaigns.
* Caps losses at the number of fleets that currently exist.
* Resolves automatically when no fleets are lost or every remaining fleet must be destroyed.
* Otherwise pauses the initiative and requires the HRAO faction to select exactly the required number of fleets.
* Creates pending decision state only when an HRAO selection is required.
* Clears the pending loss count and HRAO faction decision state after manual resolution.
* Persists fleet losses immediately, so repeated Storm at Sea events use the remaining fleet state.
* Does not prevent Rome from raising replacement fleets later.

## HRAO selection

The existing generic action form supports an accessible grouped multiselect with an exact required selection count.

Fleets are shown as checkboxes in this order:

1. Reserve fleets, ordered by fleet number.
2. Deployed fleets, grouped by campaign and ordered by fleet number within each group.

Fleet options reuse the generic game-object rendering through `object_class: "fleet"` rather than rebuilding their labels in the backend. Campaign headings reuse `Campaign.display_name` and the existing text helpers.

The action is generated only for the HRAO's faction. The standard action submission flow checks the phase, pending decision and HRAO ownership through `is_allowed()` while the game row is locked. Execution validates the exact selection count, malformed or duplicate values, and whether every selected fleet still exists in the current game.

Repeated submissions and stale or cross-game fleet selections are rejected through the normal action flow.

## Fleet Support and campaign recall

Destroying deployed fleets can leave a land campaign without its required Fleet Support.

The event records a warning when a campaign changes from supported to unsupported. It does not recall the campaign immediately, giving the Senate an opportunity to raise and assign replacement fleets.

Fleet Support is evaluated through a shared method on the existing `War` model. Both the Storm at Sea warning and the existing Senate-end recall procedure use this same calculation, avoiding a parallel support system or duplicated rules.

At the end of the Senate Phase, an unsupported campaign is automatically recalled if its Fleet Support has not been restored. A campaign reinforced during the Senate remains deployed.

## Game log

The log records:

* The occurrence of the event.
* Whether an HRAO decision is required and the number of fleets to select.
* The exact fleet numbers destroyed.
* When no fleets are lost.
* Campaigns that have newly lost the required Fleet Support.
* Any later automatic recall through the existing Senate behaviour.

Automatic destruction avoids repeating the same fleet-loss information in consecutive log entries.

Examples:

> Faction 1 drew storm at sea. The HRAO must choose 3 Roman fleets to eliminate.

> Storm at sea destroyed 2 fleets (I and IV).

> Fabius' campaign no longer has sufficient fleet support for its land battle against the 2nd Illyrian War. Unless sufficient fleet support is restored during the Senate phase, it will be automatically recalled.

The original issue draft proposed including the raw 2d6 result in the game log. Following the logging pattern used by other dice-dependent actions, this implementation records the resulting fleet losses without exposing the dice roll itself.

## Basic Game scope

This implementation deliberately affects only standard Roman Fleets.

Provincial Fleets belong to the optional Advanced Game province rules and are not used in the Basic Game. Consequently, the province system does not need to be operational for this event. Advanced Game Provincial Fleets remain out of scope.

This scope explanation is documented here rather than displayed in the in-game action interface.

## Implementation notes

* Adds the Storm at Sea event handler and HRAO resolution action.
* Adds a shared helper for fleet destruction, logging, manual pending-decision cleanup and newly unsupported campaign detection.
* Adds a shared `War.has_required_fleets()` method used by both the event and the existing Senate-end validation.
* Adds the pending fleet-loss field through migration `0064_game_storm_at_sea_fleet_losses.py`, depending on main's `0063_defeated_war_status.py` migration.
* Reuses the HRAO state already maintained by the existing game architecture instead of recalculating it during the event.
* Reuses the standard action authorization performed under the game lock instead of duplicating phase and HRAO checks inside `execute()`.
* Reuses `Campaign.display_name`, `to_sentence_case`, `pluralize` and generic object rendering.
* Preloads the campaign commander and war relationships needed to build the grouped options.
* Supports campaigns whose associated war is already marked as defeated without relying on the filtered `snapshot.wars` collection.
* Extends the generic frontend action form instead of creating an event-specific form.
* Uses stable option values as React keys for the grouped multiselect.
* Adds a focused Forum Phase preset for repeatable development and end-to-end testing.
* Updates the action documentation.
* Updates the README event checklist only to mark **Storm at Sea** as implemented.

## Review feedback addressed

The latest update incorporates ActualRory's review suggestions:

* Fleet and campaign labels now reuse the existing rendering, model properties and text helpers.
* Redundant phase, pending-state and HRAO validation has been removed from `execute()`.
* Tests now check authorization through `is_allowed()` and the normal action flow rather than depending on direct calls to redundant validation.
* Defensive cleanup of an impossible previous Storm at Sea decision has been removed.
* The event no longer calls `set_hrao()` or recalculates HRAO state.
* Automatic destruction messages and zero-loss branches have been simplified.
* The Basic Game and Provincial Fleet explanation has been removed from the player-facing `ActionDescription`.
* Tests no longer construct artificial previous pending-decision states.
* Coverage includes a campaign linked to a defeated war to prevent regression of the previous grouping `KeyError`.

## Automated testing

Coverage includes:

* The HRAO-only decision and preservation of the current initiative.
* Exact selection count.
* Malformed, duplicate, stale and cross-game selections.
* Rejection of non-HRAO and repeated submissions through the normal authorization flow.
* Reserve-first grouping and numerical fleet order.
* Generic fleet rendering.
* Campaign grouping when the associated war has defeated status.
* Reserve and deployed fleet destruction.
* Evil Omens modifying the separate effect roll.
* Automatic resolution when all remaining fleets are lost.
* Automatic resolution when no fleets exist or no losses are produced.
* Cleanup of pending decision state after manual resolution.
* Confirmation that automatic paths create no pending decision state.
* Repeated events using the persisted fleet state.
* Newly insufficient Fleet Support warnings.
* Senate reinforcement preventing recall.
* Senate-end recall when support remains insufficient.
* Frontend checkbox interaction and authorization boundaries.

Verification of the current review-feedback update:

* **mypy:** passed across 442 source files.
* **Targeted Storm at Sea and related Forum Phase tests:** 45 passed.
* **Broader Forum Phase and Fleet Support/Senate tests:** 165 passed.
* **Next.js production build and TypeScript:** passed.
* **Prettier:** passed for the changed frontend files.
* **Python compilation:** passed.

## Manual testing

The event was also exercised through complete local games:

* Multiple storms in one turn destroyed reserve and deployed fleets while preserving the remaining fleet state.
* A campaign that lost Fleet Support was reinforced during the Senate and fought normally.
* A different unsupported campaign was automatically recalled at the end of the Senate.
* Newly raised fleets were available after earlier fleets had been destroyed.
* A reserve-only scenario with no wars displayed and destroyed the correct fleets.
* A result greater than the number of remaining fleets automatically destroyed all of them.
* A further storm with no fleets resolved cleanly and allowed play to continue.
* The HRAO belonged to a non-host faction and only that faction received the action.
* Normal initiative auctions continued after the event.

The Evil Omens calculation is covered by a deterministic backend test using a roll of 5 with two Evil Omens effects, producing 3 Fleet losses. The game log reports the resulting loss count without exposing the dice roll.


backend/rorapp/models/war.py
```diff
@@ -34,6 +34,11 @@ class Status(models.TextChoices):
     fought_land_battle = models.BooleanField(default=False)
     fought_naval_battle = models.BooleanField(default=False)
 
+    def has_required_fleets(self, fleet_count: int) -> bool:
```
ActualRory, line 37: praise: Nice call putting this on the model. Sharing it with `SenatePhaseEndEffect` means the storm warning and the recall rule can't drift apart later.

backend/rorapp/actions/resolve_storm_at_sea.py
```diff
@@ -0,0 +1,174 @@
...
+                "value": fleet.id,
+                "name": f"Fleet {fleet.name}",
+                "group": (
+                    "Reserve"
+                    if fleet.campaign_id is None
+                    else campaign_group(fleet.campaign_id)
+                ),
+            }
+            for fleet in fleets
+        ]
```
ActualRory, line 80: suggestion: The frontend can build these labels for you. Other fleet multiselects send `object_class: "fleet"` with the id, and `renderObject` turns that into the same "Fleet I" string. Same story for the campaign heading: `Campaign.display_name` already gives you "Fabius' campaign", and `destroy_storm_fleets` uses it.

Going that way also lets the senator and war dicts go, and it avoids `wars[campaign.war_id]`. `snapshot.wars` leaves out defeated wars, so that lookup would raise a KeyError if a campaign's war is ever defeated while this runs.

Also worth knowing that select fields already group options through `group_by`, so the per-option `group` is a second way of doing the same job.

```suggestion
        campaigns = {campaign.id: campaign for campaign in snapshot.campaigns}

        def campaign_group(campaign_id: int) -> str:
            campaign = campaigns[campaign_id]
            return f"{to_sentence_case(campaign.display_name)} — {campaign.war.name}"

        fleets = sorted(
            snapshot.fleets,
            key=lambda fleet: (
                0 if fleet.campaign_id is None else fleet.campaign_id,
                fleet.number,
            ),
        )
        options = [
            {
                "value": fleet.id,
                "object_class": "fleet",
                "id": fleet.id,
                "group": (
                    "Reserve"
                    if fleet.campaign_id is None
                    else campaign_group(fleet.campaign_id)
                ),
            }
            for fleet in fleets
        ]
```

The import at the top needs to change with it: `to_sentence_case` instead of `possessive`.

backend/rorapp/actions/resolve_storm_at_sea.py
```diff
@@ -0,0 +1,174 @@
...
+        hrao_exists = Senator.objects.filter(
+            game=game,
+            faction=faction,
+            alive=True,
+            location="Rome",
+            titles__contains=[Senator.Title.HRAO.value],
+        ).exists()
+        if not hrao_exists:
+            return ExecutionResult(False, "Only the HRAO faction may resolve this event.")
+
```
ActualRory: suggestion: `submit_action` already calls `is_allowed` under the game row lock and raises before `execute` runs, so none of these can fail through the API. Other actions only validate the selection at this point, so this block and the HRAO query can go. The tests that reach them call `execute` directly, so they would need a tweak.

```suggestion
```

backend/rorapp/helpers/handle_event.py
```diff
@@ -44,6 +52,67 @@ def handle_event(
...
+    Faction.objects.bulk_update(factions, ["status_items"])
+
+    game.storm_at_sea_fleet_losses = fleet_losses
+    game.sub_phase = Game.SubPhase.STORM_AT_SEA
+    game.save()
+    fleet_noun = "fleet" if fleet_losses == 1 else "fleets"
+    Log.create_object(
+        game.id,
+        f"{prefix} The HRAO must choose {fleet_losses} Roman {fleet_noun} to eliminate.",
+    )
```
ActualRory: nitpick: `pluralize` in `helpers/text.py` covers this.

```suggestion
    Log.create_object(
        game.id,
        f"{prefix} The HRAO must choose {pluralize(fleet_losses, 'Roman fleet')} to eliminate.",
    )
```

Same at `resolve_storm_at_sea.py` line 160, and `ActionDescription.tsx` line 274 where `helpers/text.ts` has the matching helper.

frontend/components/ActionDescription.tsx
```diff
@@ -269,6 +270,21 @@ const ActionDescription = ({ actionName, context }: ActionDescriptionProps) => {
...
+        <p>
+          Select exactly {context.fleet_losses} Roman {fleetNoun} to eliminate.
+        </p>
+        <p className="text-sm text-neutral-600">
+          This implementation covers only the basic game. Provincial fleets
+          belong to the provincial wars advanced rule (§2.02) and are outside
+          its scope. All fleets shown here are Roman fleets.
+        </p>
+      </>
+    )
```
ActualRory, line 280: issue: Could you drop the second paragraph? No other action description cites a rule section or mentions what isn't implemented yet, and since provincial fleets don't exist in the game, it points the player at a distinction they can't see. The out of scope section of the PR description is a better home for it.

```suggestion
    return (
      <p>
        Select exactly {context.fleet_losses} Roman {fleetNoun} to eliminate.
      </p>
    )
```

backend/rorapp/helpers/handle_event.py
```diff
@@ -44,6 +52,67 @@ def handle_event(
...
+    if fleet_losses == len(fleets):
+        elimination_text = (
+            "The only existing Roman fleet must be eliminated."
+            if fleet_losses == 1
+            else f"All {fleet_losses} existing Roman fleets must be eliminated."
+        )
+        Log.create_object(
+            game.id,
+            f"{prefix} {elimination_text}",
+        )
```
ActualRory: issue: The destruction log that follows already gives the count and the fleet numbers, so this comes out as:

> Faction 1 drew storm at sea. All 3 existing Roman fleets must be eliminated.
> Storm at sea destroyed 3 fleets (I, II and III).

Could you drop the second sentence and let the destruction line carry it? The two no-loss branches can also become one, since the message reads the same whether Rome has no fleets or simply loses none.

```suggestion
    if fleet_losses == 0:
        Log.create_object(game.id, f"{prefix} No fleets were lost.")
        return True

    if fleet_losses == len(fleets):
        Log.create_object(game.id, prefix)
```

backend/rorapp/helpers/handle_event.py
```diff
@@ -44,6 +52,67 @@ def handle_event(
     return True
 
 
+def handle_storm_at_sea(
+    game: Game, current_faction: Faction, random_resolver: RandomResolver
+) -> bool:
+    clear_storm_at_sea_decision(game)
```
ActualRory: question: Is there a case where this has something to clear? A second storm can't start until the first is resolved, since the sub phase holds the initiative, and no faction holds `AWAITING_DECISION` during the forum phase. It also clears the status from every faction rather than the one holding it.

The two tests covering it set the pending state up by hand first, so they would need adjusting alongside.

```suggestion
```

backend/rorapp/helpers/handle_event.py
```diff
@@ -44,6 +52,67 @@ def handle_event(
...
+    set_hrao(game.id)
+    hrao = Senator.objects.filter(
+        game=game,
+        alive=True,
+        location="Rome",
+        faction__isnull=False,
+        titles__contains=[Senator.Title.HRAO.value],
+    ).first()
+    if hrao is None or hrao.faction_id is None:
+        raise RuntimeError("Storm at sea cannot be resolved without an HRAO faction.")
```
ActualRory: question: I don't think `set_hrao` is needed here. The title is kept current by `kill_senator` and `lay_down_command`, and the other forum phase code reads it directly.

```suggestion
    hrao = Senator.objects.filter(
        game=game,
        alive=True,
        location="Rome",
        faction__isnull=False,
        titles__contains=[Senator.Title.HRAO.value],
    ).first()
    if hrao is None or hrao.faction_id is None:
        raise RuntimeError("Storm at sea cannot be resolved without an HRAO faction.")
```

That leaves the `set_hrao` import unused. Separately, the `RuntimeError` would come back as a 500 in the middle of an initiative roll. If there's no realistic way to end up without an HRAO then the check can go, and if there is, it probably wants handling rather than a crash.

ActualRory (commented):
Great work on this one!

I ran the backend gates on 8bfab9d and everything is green: mypy clean, 637 tests passing. The test coverage here is really thorough, and I didn't spot anything that looked like a bug. The behaviour matches the event text, evil omens modifier included.

Most of my comments are about reusing bits the codebase already has. Feel free to ignore the nitpicks, but I've suggested accordingly.

Also feel free to push back if I'm mistaken on any of the below! I'm no superdev.

frontend/components/ActionDescription.tsx
```diff
@@ -231,8 +231,9 @@ const ActionDescription = ({ actionName, context }: ActionDescriptionProps) => {
   if (actionName === "Propose disbanding forces") {
     return (
       <p>
-        Disbanding a legion or fleet saves the State 2T per turn in maintenance. 
-        You may only disband reserve forces, and cannot disband forces raised this turn.
+        Disbanding a legion or fleet saves the State 2T per turn in maintenance.
+        You may only disband reserve forces, and cannot disband forces raised
+        this turn.
```
ActualRory, line 237: nitpick: These lines are rewrapped but otherwise unchanged. Same with the blank line near `GenericActionForm.tsx` 236 and the dependency array at 381. Does this do anything visually?

ActualRory: Actually eh, this is probably for a nice wrapping reason and I presume you've tested how it looks. If so, no worries!

backend/rorapp/helpers/handle_event.py
SoyMakari: Thanks for the detailed review, @ActualRory. I’ve pushed an update addressing your suggestions: fleet and campaign labels now reuse the existing rendering/model helpers, the redundant validation, defensive pending-state cleanup and set_hrao() call have been removed, and the automatic logs and UI description have been simplified. I’ve also updated the tests to follow the real action flow and added coverage for campaigns linked to defeated wars. The relevant backend suite (165 tests), mypy and the frontend build all pass locally. The GitHub checks are currently still queued awaiting runners. When you have a chance, could you please take another look?

ActualRory (approved):

ActualRory:
Accidentally made a commit via the UI, reverted it, but now you'll need to self-approve @SoyMakari - good PR though, ready to merge! Thanks

SoyMakari:
Thanks! @ActualRory  I'm not entirely sure what you mean by “self-approve” here — I may be missing something, but GitHub doesn't seem to let me approve my own PR.

Do you mean that someone else needs to approve the latest push because it was made from your account?

ActualRory:
> Thanks! @ActualRory I'm not entirely sure what you mean by “self-approve” here — I may be missing something, but GitHub doesn't seem to let me approve my own PR.
> 
> Do you mean that someone else needs to approve the latest push because it was made from your account?

Yep! My bad. If you make a meaningless single character change (and perhaps even change it right back) that'll make me able to approve again, otherwise Logan can scoop it up.

ActualRory (approved):

ActualRory:
I'm mistaken, needs Logan's approval now. Mea culpa!

iamlogand (approved):
Approving on the basis that @ActualRory has already approved

PR #836, opened 2026-09-10
Capitalize concession names in the UI (Fixes #819)

### LINKED ISSUE #819: Better capitalisation for concessions
Image

When we do concessions, we should probably tidy up capitalisation for the actual concession names. 

Super low priority, of course.

SoyMakari:
Fixes #819

## Summary

* Render assigned concession names in `SenatorDisplay` with the existing `toSentenceCase()` helper.
* Replace the remaining frontend uses of `first-letter:uppercase` with display-time `toSentenceCase()` calls.
* Keep the underlying values unchanged for game logic, comparisons and lookups.
* Remove the UI test and its dedicated preset following maintainer feedback.

## Details

Assigned concession names relied on Tailwind's `first-letter:uppercase` utility, which did not capitalize the displayed text reliably.

Using the existing `toSentenceCase()` helper makes the capitalization explicit and consistent. For example:

* `ship building` → `Ship building`
* `harbor fees` → `Harbor fees`
* `Aegyptian grain` remains `Aegyptian grain`

Following the review feedback, the remaining uses of `first-letter:uppercase` in the frontend were also replaced.

The affected display labels include:

* senator titles, concessions and statuses;
* faction card labels;
* unawarded concession names;
* faction and war statuses;
* game phase and game status.

The transformation is applied only when rendering text. Original values continue to be used unchanged for comparisons, lookups and gameplay logic.

No new helper, component, backend behavior or UI test has been introduced.

## Validation

* Manually verified the concession capitalization in the locally running application.
* Backend CI passed on the current PR head.
* Frontend CI passed on the current PR head.
* No new UI tests were added, in line with the maintainer's review guidance.

## Screenshot

image


frontend/tests/concessionCapitalization.spec.ts
iamlogand, line 1: issue: this will be unfortunate news if you put a lot of effort into this test, but we're avoiding any new UI tests for now due to a planned UI overhaul after the early republic features are done. I understand that you are not able to run the app locally, so it makes sense that you needed to use a UI test to validate that your fix works. Would you be able to remove this test?

frontend/components/SenatorDisplay.tsx
```diff
@@ -50,16 +50,16 @@ const SenatorDisplay = ({ senator }: SenatorDisplayProps) => {
                           <Popover
                             className="flex"
                             trigger={
-                              <span className="text-yellow-900 first-letter:uppercase">
```
iamlogand, line 53: suggestion: It looks like `first-letter:uppercase` doesn't actually work, but we are using it in a few other places as well. Could you replace all of them?

I don't think you need to test each one because the change is small

iamlogand (dismissed):

frontend/tests/concessionCapitalization.spec.ts
SoyMakari: Thanks Logan. I can run the app locally now — the screenshot in the PR is from a manual local test. I added the UI test as additional coverage rather than because I couldn't run the app.

I've removed the UI test and its dedicated preset as requested.

frontend/components/SenatorDisplay.tsx
SoyMakari: Done — I replaced all the remaining `first-letter:uppercase` uses in the frontend with the existing `toSentenceCase()` helper. The changes are display-only and the underlying values are still used unchanged for comparisons and lookups.

I haven't added any new UI tests.

ActualRory (approved):
Good fix!

PR #839, opened 2026-09-11
cosmetic: Always show unaligned senators (Closes #741)

### LINKED ISSUE #741: Make the unaligned senators section always visible
This prevents a bad usability scenario where the user is looking for the unaligned senators but can't find them. If there is always an unaligned senators section, there will never be confusion about where it is.

When there are no unaligned senators, it can just say "There are no unaligned senators right now" or similar.

SoyMakari:
Fixes #741

## Summary

* Always display the `Unaligned senators` section.
* Keep rendering existing unaligned senators through `SenatorDisplay`.
* Show an empty-state message when there are currently no unaligned senators.
* Preserve the existing deceased-senator popover and families-that-may-return behavior.

## Details

Previously, the entire `Unaligned senators` section was hidden when there were no living unaligned senators and no deceased senators.

This could make it unclear where unaligned senators would normally appear.

The section is now always visible. When the list is empty, it displays:

> There are no unaligned senators right now

When unaligned senators are present, they continue to use the existing filtering, sorting and `SenatorDisplay` rendering.

The deceased-senator popover remains unchanged.

## Validation

* Manually verified in the locally running application with a game starting with no unaligned senators.
* Confirmed that the section remains visible and displays the empty-state message.
* No new UI tests were added, in line with the maintainer's current guidance on UI testing.

## Screenshot

Begining of the game

image

Persuasion attempt

image


frontend/components/GameMain.tsx
```diff
@@ -173,44 +170,46 @@ const GameMain = ({ publicGameState, privateGameState }: Props) => {
...
-      )}
+        <div className="grid grid-cols-[repeat(auto-fill,minmax(700px,1fr))] gap-4">
+          <div className="rounded border border-neutral-400">
+            <div className="divide-y divide-neutral-300 py-0.5">
+              {unalignedSenators.length > 0 ? (
+                unalignedSenators.map((senator: Senator, index: number) => (
+                  <SenatorDisplay key={index} senator={senator} />
+                ))
+              ) : (
+                <div className="flex items-center py-2 pl-3 pr-4 text-neutral-600 lg:pl-5 lg:pr-6">
```
ActualRory: `flex items-center` has no effect here: one text child, and the height is driven by the content.

```suggestion
                <div className="py-2 pl-3 pr-4 text-neutral-600 lg:pl-5 lg:pr-6">
```

ActualRory (commented):
Looks good, thank you! One suggestion inline.

frontend/components/GameMain.tsx
ActualRory: Total nitpick! But only a few-click fix now that I've written the suggestion if you want to run with it. Shout if I'm wrong - this is 1AM and two-beer PR reviewing.

ActualRory (approved):

PR #843, opened 2026-09-14
feat: add famine profiteering on grain concessions (§1.06.12)

ActualRory:
Adds famine profiteering. While a famine is in effect, a senator holding a grain concession may take extra income from it at the cost of popularity (§1.06.12 and the Aegyptian and Sicilian grain concession cards).

> Grain Concession bonuses may be taken for Drought even if the Senator's Popularity is -9. (1.06.12)

> Voluntary x2 Grain Concession Income and −2 Popularity if Drought in effect. +1 Income multiplier and −1 Popularity per additional Drought effect. (Aegyptian Grain, Sicilian Grain)

Ruled-as-written, this decision would take place during the actual calculation of income for your senators. I've done it as an action during redistribution instead. This avoids adding a new pause to the game every time income is calculated. This has zero effect on actual gameplay/strategy.

### Behaviour

- A new **'Profiteer from famine'** action is offered during redistribution to any faction with a senator holding a grain concession, while famine severity is above 0.
- The player picks the concession, and each option shows what it pays and costs, e.g. `Sicilian grain (+4T, -2 popularity)`.
- At famine severity 'X' the concession pays X extra multiples of its income and the senator loses X + 1 popularity. The revenue step has already paid the base income, so only the extra is added.
- Each grain concession can be used once per revenue phase. The senator carries a `profiteered from <concession>` status until redistribution ends.
- The bonus can still be taken at -9 popularity. The loss is clamped, and the popularity sentence is dropped from the log when nothing was lost.

### New log examples

```
Senators in Faction 1 earned 8T of revenue.
Senators in Faction 2 earned 3T of revenue.
Senators in Faction 3 earned 3T of revenue.
Cornelius profiteered from the famine on the Sicilian grain concession, earning an extra 4T. He lost 2 popularity.
```

At -9 popularity:

```
Cornelius profiteered from the famine on the Sicilian grain concession, earning an extra 4T.
```

### Implementation notes

- As I mentioned above, the choice is offered in the existing redistribution sub-phase rather than a new step before revenue. Nothing between revenue collection and redistribution reads a senator's talents or popularity, so the outcome is the same as deciding at collection.
- `GRAIN_CONCESSION_REVENUE` in `classes/concession.py` now holds the 5T and 4T figures, so `RevenueEffect` and the new action share them.
- The status item follows the existing `bribed {n}T` pattern (a string built by a classmethod on `Senator.StatusItem`), so a senator holding both grain concessions can decide on each one separately.
- `game.famine_severity` already counts undefeated famine wars plus drought events, which is what "Drought in effect" means. Drought events expire at the start of the forum phase, so a drought drawn in the forum phase is still in effect in the next turn's revenue phase.

### Testing

- `mypy .`: no issues in 441 source files
- `pytest`: 632 passed
- New `s1_06_12_famine_profiteering_test.py`, 6 cases: income and popularity at severity 1 and 2, allowed at -9 popularity, once per concession, offered only during a famine.


backend/rorapp/actions/profiteer_from_famine.py
```diff
@@ -0,0 +1,111 @@
...
+                if c.value == selection["Concession"]
+            ),
+            None,
+        )
+        if claim is None:
+            return ExecutionResult(False)
+        senator, concession = claim
+
+        # Revenue already paid the base income, so only the extra multiples are added (1.06.12)
+        severity = Game.objects.get(id=game_id).famine_severity
```
iamlogand: suggestion: You could create `GameStateLive` once at the start of `execute` and use it for passing to `_unclaimed_grain_concessions` and for getting severity

backend/rorapp/effects/redistribution_done.py
```diff
@@ -22,11 +22,13 @@ def execute(self, game_id: int, random_resolver: RandomResolver) -> bool:
...
 
-        # Remove contributed status
+        # Remove contributed and profiteered statuses
         senators = Senator.objects.filter(game=game_id)
         for senator in senators:
-            if senator.has_status_item(Senator.StatusItem.CONTRIBUTED):
-                senator.remove_status_item(Senator.StatusItem.CONTRIBUTED)
+            senator.remove_status_item(Senator.StatusItem.CONTRIBUTED)
+            senator.status_items = [
+                s for s in senator.status_items if not s.startswith("profiteered")
```
iamlogand: suggestion: Since this is a magic string, you could define a `PROFITEERED_PREFIX` in the `Senator.StatusItem` class and add a helper method to the senator class, like we already have for bribes so you don't even need to access the prefix outside of the senator class.

Side note: I just noticed we're using "bribe" as a magic string within the senator class, which isn't ideal, but don't fix that here.

iamlogand (dismissed):
Nice! These code quality suggestions are non-blocking

backend/rorapp/actions/profiteer_from_famine.py
ActualRory: Done in 42229cd! Though `GameStateLive.game` seems to run `Game.objects.get` on each access, so this is tidier but it might still be making the call?

backend/rorapp/effects/redistribution_done.py
ActualRory: Done in dfa7614!

`has_profiteered`, `add_profiteered` and `clear_profiteered` sit beside the bribe helpers, and the prefix string is now only in the model. It's three methods rather than a get/set pair because a senator can hold both grain concessions, so set_x(None) does not map onto clearing one of them - hope that makes sense!

iamlogand (approved):

PR #850, opened 2026-09-16
feat: enable use of feature flags

iamlogand:
Adds feature flag infrastructure to support trunk-based development. Flags are read automatically from `FEATURE_FLAG_<NAME>` environment variables, requiring no settings.py changes per flag. Includes the `civil_war` flag as the first consumer, gating `LayDownCommandEffect`'s auto-behaviour so the game remains playable with the flag off whilst the civil war feature is developed.


ActualRory (approved):

ActualRory:
Looks good! Gonna speedmerge this to unblock the civil war bits.

PR #851, opened 2026-09-16
feat: discord invite link

### LINKED ISSUE #830: Add Discord server invite link to game list page
There is a Discord server for this project. It has a looking-for-games text channel that could be used by players to find others to play with.

## Security note

The actual URL shouldn't be committed, because that would allow the URL to be found without authenticating at roronline.com. It should be stored as an environment variable in the backend, exposed to authenticated users over the API, and conditionally rendered at the top of the game list page.

iamlogand:
Closes #830 by adding a discord invite link

## Screenshot

image


backend/rorapp/views/discord_invite_url.py
```diff
@@ -0,0 +1,16 @@
+import os
+from django.http import JsonResponse
+from dotenv import load_dotenv
+from pathlib import Path
+from rest_framework.decorators import api_view, permission_classes
+from rest_framework.permissions import IsAuthenticated
+
+BASE_DIR = Path(__file__).resolve().parent.parent
+load_dotenv(os.path.join(BASE_DIR, ".env"))
+discord_invite_url = os.getenv("DISCORD_INVITE_URL")
```
ActualRory, line 10: This won't be set in production yet. Right? It works locally only because backend/.env is loaded, but I'm not sure where you're sorting that prod-side.

deploy_backend.yml only replaces ${TAG}, and the container gets every other variable through the ${VAR} entries in backend/docker-compose.yml. `DISCORD_INVITE_URL` isn't listed there, so os.getenv returns None on Elastic Beanstalk and the link never shows. 

I'd put `DISCORD_INVITE_URL: ${DISCORD_INVITE_URL}` under app.environment in that file, plus the value set as an EB environment property.

Could also add DISCORD_INVITE_URL= (empty, without a real URL) to the .env template in docs/setup-development-environment.md, so others know the link is controlled by a variable.

Let me know if I'm wrong here!

ActualRory (changes requested):

backend/rorapp/views/discord_invite_url.py
iamlogand: I added it to EB as an environment variable, but completely forgot about containerization! I've added an environment entry to docker-compose.yml so it'll get pulled into the container. I've also updated the docs to match.

ActualRory (approved):
Looks good now! Let's send it to prod and we can revert if not - it won't be breaking either way.

ActualRory:
It's live and works! Nice.

PR #852, opened 2026-09-16
feat: add Declaration of Civil War behind the civil_war flag (§1.11.3)

ActualRory:
Split out of #816 at review: the Declaration of Civil War and the choice to lay down command, without rolling for legions or the civil war battle. Both actions sit behind the `civil_war` flag from #850.

> Starting with the HRAO and proceeding clockwise, each Commander who gained a Land Victory in this turn (including a Senate Victory) must either lay down his command (i.e., return his Force to the Active Forces boxes and return to Rome) or declare himself in Revolt. [...] Only one Faction may be in Revolt at a time. Should a Faction that wishes to Revolt field a stronger Army (inclusive of the respective Commanders' Military ratings) than that of a previously declared Rebel, the initial Rebel's declaration is ignored [...] Fleets play no role in a Civil War and any Fleets that started with the Commander immediately return to the Active Forces boxes upon declaration. - §1.11.3

### Behaviour

- With the flag off, every land victor lays down his command automatically, exactly as on main.
- With the flag on, land victors decide in turn starting with the HRAO's faction, choosing **Lay down command** or **Declare civil war**.
- A declaration creates the Civil War, gives the commander a rebel marker, moves him to Italia and sends his fleets back to the reserve.
- A second faction is offered the declaration only when its army is strictly stronger. The earlier declaration is then ignored and that commander lays down his command. Once the phase ends, nobody else may revolt while the Primary Rebel stands.
- The Senate cannot recall, reinforce or replace the rebel, and he does not become a Proconsul. The UI marks him "Rebel" and his war and army "Marching on Rome".

### New log examples

```
Cornelius declared himself in revolt and marched on Rome with 5 legions (I–V).
Manlius fielded the stronger army, so Cornelius' declaration was ignored.
```

### Implementation notes

- The Civil War is an ordinary `War` row with `primary_rebel` set, so the state debit, the four-active-wars loss, unprosecuted-war unrest and the deploy proposal need no new code.
- The flag is checked only in the two actions' `is_allowed`. `LayDownCommandEffect` stays as #850 left it and remains the flag-off path.
- `helpers/civil_war.py` reads only the game state it is given, so `is_allowed` makes no queries (0 captured against a snapshot). `helpers/declare_civil_war.py` is the one that writes.
- Declaration order reads the HRAO title.
- A `declared revolt` status item marks a rebel who declared this phase, since only he can be displaced. It is cleared when the phase ends.

### Out of scope

Rolling for legions (§1.11.31), the civil war battle (§1.11.36 to §1.11.373) and a revolt failing on the rebel's death stay on #816, which will be rebuilt on top of this. Secondary rebels, markers, maintenance and winning conditions come later. README box stays unticked until the flag is removed.

### Testing

662 backend tests pass and `mypy .` is clean across 450 files. The lay down tests run once with the flag off and once with it on, so we know feature flags should be working properly.


backend/rorapp/actions/lay_down_command.py
iamlogand, line 1: note: This approach allows one faction to lay down command multiple times. This is a crazy edge case though, so don't even attempt to improve it. I'll document this in an issue so future us can one day deal with it.

I think a good way to handle this (in the far future) will be to let the player choose which of their victorious commanders will rebel, allowing both commanders to roll for legions before making this decision.

The crazy edge cases that come up in this adaptation... 😅

backend/rorapp/helpers/civil_war.py
```diff
@@ -0,0 +1,76 @@
...
+    game = game_state.game
+    if (
+        game.phase != Game.Phase.REVOLUTION
+        or game.sub_phase != Game.SubPhase.CIVIL_WAR_DECLARATION
+    ):
+        return None
+    victors = land_victors_in_declaration_order(game_state)
+    if victors and victors[0].commander and victors[0].commander.faction_id == faction_id:
+        return victors[0]
+    return None
```
iamlogand: note: You've set this up quite differently to how I would have done. This isn't a criticism.

During every step of `CIVIL_WAR_DECLARATION`, for each faction you determine the declaring order. Instead of doing that I'd probably have done this:
- Add a new effect that fires during `CIVIL_WAR_DECLARATION` whenever nobody has `AWAITING_DECISION`
- The new effect would have one job: find the next declaring campaign faction and give them the `AWAITING_DECISION` status item.
- Then I'd gate `DeclareCivilWar` and `LayDownCommand` on the faction having `AWAITING_DECISION`.

Happy for you to do it your way, but thought it would be worth making you aware of this alternative approach in case you hadn't considered it. The benefit of your approach is it doesn't require adding and removing the `AWAITING_DECISION` faction status item, but it would only happen a couple times per turn.

backend/rorapp/helpers/declare_revolt.py
iamlogand, line 1: ~idea: Wouldn't it be cleaner to split the declaration into two stages?~

~At declaration time, commit only what the displacement check needs, i.e. add `DECLARED_REVOLT` to the senator and create/update the War row with `land_strength` and `primary_rebel`. Then defer the major state changes (e.g. sending fleets to reserve, dismissing the master of horse, moving the commander to Italia, setting `rebel = True`) to phase end once the Primary Rebel is confirmed. That way, if the standing rebel is displaced there's nothing to undo, his `DECLARED_REVOLT` is cleared and `War.primary_rebel` is updated to the new rebel, but his legions, fleets and MoH were never touched, so `lay_down_command` can run as normal.~

Ignore this - it would be more hassle than it's worth, your current approach is good

iamlogand (commented):
Some thoughts

backend/rorapp/helpers/declare_civil_war.py
iamlogand: Actually, I'm not entirely sure it would be cleaner. Maybe your current approach is simpler. Will come to a proper conclusion tomorrow.

backend/rorapp/actions/lay_down_command.py
ActualRory: Lmao fair, I'll take that offer and leave it as-is -we can always refactor later if this produces a problem!

backend/rorapp/actions/declare_revolt.py
iamlogand, line 1: suggestion: One thing that's bothering me is that we're using the terms "civil war" and "revolt" interchangably. This stems from the fact that the living rules uses these terms interchangably.

I propose that we be consistent in terminology throughout the codebase and any player-visible text:
- Revolt - the term we use for the act of declaration and the resulting war
- Rebel - the term we use for a senator or force that's part of the revolt

Would you mind renaming every occurance of civil war to revolt?

Two things that are already called civil war before this PR:
- CIVIL_WAR_DECLARATION status item - could you rename this?
- "civil_war" feature flag - let's just leave this as-is, since it's temporary

If you do this, I'll update terminology-changes.md to mention this

backend/rorapp/helpers/declare_revolt.py
```diff
@@ -0,0 +1,89 @@
...
+    # The Master of Horse keeps his office and returns to Rome (1.11.32)
+    master_of_horse = campaign.master_of_horse
+    if master_of_horse:
+        master_of_horse.location = "Rome"
+        master_of_horse.save()
+        campaign.master_of_horse = None
+
+    legions = list(campaign.legions.all().order_by("number"))
+    log_text = f"{commander.display_name} declared himself in revolt and marched on Rome with "
+    log_text += unit_list_to_string(legions, []) if legions else "no legions"
```
iamlogand: note: I didn't realise it was possible to declare revolt, march on Rome, and win, all with no legions - but it is 😆

In fact, if the senate were to deploy 3 legions and a 1 MIL commander to put down the revolt, there is a 0.5% chance that the rebel still wins the battle and therefore the game!

backend/rorapp/helpers/civil_war.py
```diff
@@ -0,0 +1,76 @@
...
+from rorapp.game_state.game_state_live import GameStateLive
+from rorapp.game_state.game_state_snapshot import GameStateSnapshot
+from rorapp.models import Campaign, Game, Legion, Senator
+
+
+def army_strength(legions: List[Legion], military: int) -> int:
+    """Strength of an army, with the commander's Military rating capped by it (1.11.37)."""
+
+    legion_strength = sum(l.strength for l in legions)
+    return legion_strength + min(military, legion_strength)
```
iamlogand: suggestion: `resolve_combat.py` has this exact logic already for calculating army strength. Could you change that code to call this helper so we're not repeating the logic? If you decide to do that, it would also make sense to move this helper to a different file.

backend/rorapp/helpers/civil_war.py
iamlogand: issue: Your approach leaves no way for other players or spectators to see who is currently deciding. In the current UI, I'm talking about the "Awaiting decision" faction pill, and in the future this could be something more user-friendly, but either way the other clients would need some way to detect the status). For this reason, I'd like you to change to the above approach.

backend/rorapp/helpers/declare_civil_war.py
```diff
@@ -0,0 +1,89 @@
...
+        .first()
+    )
+    displaced_rebel = displaced_war.primary_rebel if displaced_war else None
+    if displaced_war and displaced_rebel:
+        displaced_rebel.rebel = False
+        displaced_rebel.remove_status_item(Senator.StatusItem.DECLARED_REVOLT)
+        displaced_rebel.save()
+        Log.create_object(
+            game_id,
+            f"{commander.display_name} fielded the stronger army, so "
```
iamlogand: nitpick

```suggestion
            f"{commander.display_name} fielded the stronger rebel army, so "
```

backend/rorapp/helpers/declare_civil_war.py
```diff
@@ -0,0 +1,89 @@
...
+            f"{possessive(displaced_rebel.display_name)} declaration was ignored.",
+        )
+        for displaced_campaign in Campaign.objects.filter(
+            game=game_id, war=displaced_war
+        ).order_by("id"):
+            lay_down_command(displaced_campaign)
```
iamlogand: question: is the loop here in preparation for secondary rebels? The code as is suggests we'd only have one displaced campaign.

backend/rorapp/helpers/declare_civil_war.py
```diff
@@ -0,0 +1,89 @@
...
+
+    # The rebel has left Rome, so Rome needs a new highest official (1.09.11)
+    set_hrao(game_id)
```
iamlogand: issue: This commant is misleading, and the `set_hrao()` call is redundant, because the rebel can't possibly be HRAO (if they can then that's a separate bug)

backend/rorapp/helpers/declare_civil_war.py
```diff
@@ -0,0 +1,89 @@
...
+        fleet_support=0,
+        naval_strength=0,
+        spoils=0,
+        location=CIVIL_WAR_LOCATION,
+        status=War.Status.ACTIVE,
+    )
+    civil_war.primary_rebel = commander
+    civil_war.land_strength = army_strength(legions, commander.military)
+    civil_war.save()
+    campaign.war = civil_war
```
iamlogand: praise: I like that you redirect the existing campaign object that fought for Rome to march against Rome.

Campaigns are objects that I can't imagine anyone would need to inspect after they are over (within the context of a future UI that might support inspection of things like dead senators, defeated wars, etc), so this is safe to do!

backend/rorapp/helpers/declare_revolt.py
```diff
@@ -0,0 +1,89 @@
...
+    log_text += "."
+    if master_of_horse:
+        log_text += f" {master_of_horse.display_name} returned to Rome."
+    if fleets:
+        log_text += (
+            f" {unit_list_to_string([], fleets)} played no part in the revolt and "
+            "returned to the reserve forces."
+        )
+    Log.create_object(game_id, log_text)
```
iamlogand: praise: this log looks perfect 👍

frontend/components/GameMain.tsx
```diff
@@ -544,11 +544,13 @@ const GameMain = ({ publicGameState, privateGameState }: Props) => {
...
-                                war.navalStrength === 0 ? "land" : "naval"
-                              } battle`
-                            : "Awaiting the revolution phase to lay down command"}
+                          {commander?.rebel
+                            ? "Marching on Rome"
+                            : war
+                              ? `Preparing for a ${
+                                  war.navalStrength === 0 ? "land" : "naval"
+                                } battle`
+                              : "Awaiting the revolution phase to lay down command"}
```
iamlogand: suggestion: To improve readability could you refactor this to avoid the nested ternary chain? For example you could declare a new variable near where `recallReason` is declared, build the text there, then conditionally render it here.

frontend/components/GameMain.tsx
```diff
@@ -544,11 +544,13 @@ const GameMain = ({ publicGameState, privateGameState }: Props) => {
                         </p>
                       ) : (
                         <p className="text-sm text-neutral-600">
-                          {war
-                            ? `Preparing for a ${
-                                war.navalStrength === 0 ? "land" : "naval"
-                              } battle`
-                            : "Awaiting the revolution phase to lay down command"}
+                          {commander?.rebel
+                            ? "Marching on Rome"
```
iamlogand: praise: The new text is certainly useful for players to see

iamlogand (changes requested):
Great progress here! Thanks for splitting it up. This PR is just about small enough for me to review it effectively.

I've suggested a significant change in approach for how the system tracks who is currently deciding whether to declare revolt. Apart from that, I've just requested small things that shouldn't be too much trouble.

I had some doubts about the approach of declaring revolt then rolling back changes if it gets replaced, but ultimately your approach is sound and the alternative idea I had would have been more complicated.

backend/rorapp/helpers/civil_war.py
ActualRory: Very good catch!! I'll fix this post-haste, tally-ho (probably tonight)

iamlogand: No rush, of course!

backend/rorapp/helpers/declare_revolt.py
ActualRory: This is an edge case I didn't know was possible until putting this code together either! All I could think of was that scene from Emperor's New Groove. The one where Kronk goes "Well you got me, by all accounts it doesn't make sense." But with a diagram of the board state on the little board he pulls down, showing Caesar crossing the rubicon alone.

If I was artsy, I'd put it together 😂

ActualRory: It must have occurred in at least one game, but sadly not any of mine - yet!

ActualRory:
> I've suggested a significant change in approach for how the system tracks who is currently deciding whether to declare revolt. Apart from that, I've just requested small things that shouldn't be too much trouble.
> 
> I had some doubts about the approach of declaring revolt then rolling back changes if it gets replaced, but ultimately your approach is sound and the alternative idea I had would have been more complicated.

Implementing all the requested changes now. As for the alternative approach, I wouldn't mind refactoring this if you don't mind reviewing the refactored code. We can also mark it as a todo once the game is feature complete - I'll probably have a surplus code time once you move onto assets.

backend/rorapp/helpers/civil_war.py
ActualRory: Done! `RevoltDeclarationNextEffect` hands `awaiting decision` to the next land victor's faction whenever nobody holds it. Both actions are gated on it

backend/rorapp/helpers/declare_civil_war.py
ActualRory: In the immortal words of the great warrior-poet Sean Connery, there can be only one. Replaced the loop with `displaced_war.campaigns.get()`.

backend/rorapp/helpers/declare_civil_war.py
ActualRory: Comment was indeed misleading, I've now corrected it to:     
`# Senators returning to Rome may outrank the HRAO (1.09.11)`

I've also improved the logic for reasons I'll get to below.

Essentially, this was a bit of a caveman solution to guard against the below edge case:

- Dictator rebels.
- Proconsul with a larger army rebels.
- Dictator has the smaller army, so must return to Rome.
- Dictator is now HRAO.

This is also true if Rome has been gutted of office-holders, as the Master of Horse being sent home by a rebelling dictator could conceivably become HRAO.

With that in mind, it doesn't need to be as broad/redundant as I had it here. I've made this check only run if there's a displaced rebel of master of horse returning to Rome. Seems to work nicely and is a bit more explicit in its logic now.

Hope that makes sense!

iamlogand:
> > I've suggested a significant change in approach for how the system tracks who is currently deciding whether to declare revolt. Apart from that, I've just requested small things that shouldn't be too much trouble.
> > 
> > I had some doubts about the approach of declaring revolt then rolling back changes if it gets replaced, but ultimately your approach is sound and the alternative idea I had would have been more complicated.
> 
> Implementing all the requested changes now. As for the alternative approach, I wouldn't mind refactoring this if you don't mind reviewing the refactored code. We can also mark it as a todo once the game is feature complete - I'll probably have a surplus code time once you move onto assets.

I wouldn't mind reviewing after the refactor. I've spent quite a while getting familiar with this PR so it shouldn't take that long to review even after the refactor. Also I'd kinda prefer for this to be solved sooner rather than later.

backend/rorapp/helpers/declare_civil_war.py
ActualRory: Thanks! It's similar to how by how my group usually just slides everything off the war card in one-go to setup the rebellion. This is basically the code version of that!

frontend/components/GameMain.tsx
ActualRory: It's now built as `campaignStatus` next to `recallReason`. Does that work?

backend/rorapp/helpers/civil_war.py
ActualRory: Done. Both the land and naval branches of `resolve_combat` now call it. It's `force_strength(force, military)` in `helpers/force_strength.py` so the fleet count can use it too.

backend/rorapp/actions/declare_revolt.py
ActualRory: Done. I've included the sub-phase (`revolt declaration`), the action (`Declare revolt`), the helpers and the war name (`Revolt`) - I think that's everything.

ActualRory: Except the `civil_war` feature flag, which I've left in all its temporary glory.

ActualRory:
> > > I've suggested a significant change in approach for how the system tracks who is currently deciding whether to declare revolt. Apart from that, I've just requested small things that shouldn't be too much trouble.
> > > I had some doubts about the approach of declaring revolt then rolling back changes if it gets replaced, but ultimately your approach is sound and the alternative idea I had would have been more complicated.
> > 
> > 
> > Implementing all the requested changes now. As for the alternative approach, I wouldn't mind refactoring this if you don't mind reviewing the refactored code. We can also mark it as a todo once the game is feature complete - I'll probably have a surplus code time once you move onto assets.
> 
> I wouldn't mind reviewing after the refactor. I've spent quite a while getting familiar with this PR so it shouldn't take that long to review even after the refactor. Also I'd kinda prefer for this to be solved sooner rather than later.

Just as a sanity-check, we're talking about this one - right? I've been working on it but want to be sure I'm not running in the wrong direction.

> At declaration time, commit only what the displacement check needs, i.e. add DECLARED_REVOLT to the senator and create/update the War row with land_strength and primary_rebel. Then defer the major state changes (e.g. sending fleets to reserve, dismissing the master of horse, moving the commander to Italia, setting rebel = True) to phase end once the Primary Rebel is confirmed. That way, if the standing rebel is displaced there's nothing to undo, his DECLARED_REVOLT is cleared and War.primary_rebel is updated to the new rebel, but his legions, fleets and MoH were never touched, so lay_down_command can run as normal.

ActualRory:
I ran ahead with that, done. I'll look on it with fresh eyes tomorrow eve if you don't beat me to it.

iamlogand:
@ActualRory I should have been clearer - I meant this one https://github.com/iamlogand/republic-of-rome-online/pull/852#discussion_r4040834664! 😮

backend/rorapp/effects/revolution_phase_end.py
```diff
@@ -21,6 +23,16 @@ def execute(self, game_id: int, random_resolver: RandomResolver) -> bool:
             faction.remove_status_item(FactionStatusItem.DONE)
         Faction.objects.bulk_update(factions, ["status_items"])
 
+        senators = list(Senator.objects.filter(game=game_id, rebel=True))
+        for senator in senators:
+            if senator.has_status_item(Senator.StatusItem.DECLARED_REVOLT):
+                march_on_rome(Campaign.objects.get(game=game_id, commander=senator))
+            senator.remove_status_item(Senator.StatusItem.DECLARED_REVOLT)
+            # Once the rebels are determined, each loses his offices (1.11.33)
+            for office in MAJOR_OFFICES:
```
iamlogand, line 32: issue (non-blocking): this is redundant behond the first revolution phase that a revolt is declared. Maybe gate all of this new code behind a check to see if any senators have `DECLARED_REVOLT`

backend/rorapp/models/war.py
iamlogand, line 20: issue (non-blocking): we now violate this validator. However, we never actually use these validators AFAIK. I might remove them.

iamlogand (approved):
Good to merge!

I'll raise separate issues for the two small things I found.

PR #870, opened 2026-09-19
feat: fail a revolt when its Primary Rebel dies (§1.11.372)

ActualRory:
A revolt now fails when its Primary Rebel dies, however he dies (§1.11.372). Until now a rebel who died in the Mortality Phase left an active Revolt war behind him - no label for 'oversight' so I'm calling it a feature.

> If the Primary Rebel is killed in battle ([1.10.7]) or through a Mortality Chit draw ([1.05.21]), the Revolt fails. When a Revolt fails, all Secondary Rebels are killed, any surviving Rebel Forces [...] are returned to Senate control (i.e., place them in the Active Forces boxes), and any Senate Armies that had not achieved a "Victory" but were still facing Rebels return to Rome. - §1.11.372

### Behaviour

- Any death of the Primary Rebel ends the revolt: mortality, an epidemic, the mob, assassination or execution.
- The Revolt war is marked defeated rather than deleted.
- His surviving legions return to the reserve.
- Any Senate army sent against him returns to Rome. Its commander loses his Proconsul title and its forces return to the reserve.
- A rebel faction leader's heir is not a rebel, so the heir's faction may revolt again later.

### New log examples

```
Cornelius of Faction 1 died of natural causes.
The revolt ended with the death of Cornelius.
5 legions (I–V) returned to the reserve forces.
Manlius returned to Rome. 1 legion (VI) returned to the reserve forces.
```

### Implementation notes

- `kill_senator` looks up the rebel's revolt before he dies and calls `fail_revolt` once his death is logged. `fail_revolt` in `helpers/fail_revolt.py` runs every campaign on the Revolt war through the existing `lay_down_command`, then recalculates the HRAO, since a returning commander may outrank him.
- The war is saved with `update_fields`. A rebel without a family card has his row deleted, and `primary_rebel` is `SET_NULL`, so a full save would write his id back.
- **Question on the flag:** there is no `civil_war` check here, because no revolt can exist with the flag off. I'd like to gate only the entry points (the actions, and any effect that could fire with no rebel in play) for the rest of the revolt PRs, rather than every effect that only acts on a rebel. Does that work for you?

### Out of scope

Killing the Secondary Rebels when the revolt fails comes with Secondary Rebels (§1.11.32). The return of Rebel Governors' garrisons waits on Rebel Governors (§2.03). The revolt battle, including the rebel dying in it, is its own PR.

### Testing

678 backend tests pass and `mypy .` is clean across 457 files. 4 new tests in `s1_11_372_revolt_fails_test.py`; each fails with its fix removed.


iamlogand (approved):
You meant to introduce this separately 😉

ActualRory:
> You meant to introduce this separately 😉

Bingo, top-tier atomisation for easy reviewing

PR #871, opened 2026-09-19
refactor: move the Combat Results Table out of resolve_combat (§1.10.11)

ActualRory:
Moves the Combat Results Table (§1.10.11) out of `resolve_combat` into `helpers/combat_results.py`, ahead of the revolt battle (§1.11.37).

> Any combat losses called for by the Combat Results Table are applied to both armies, except, in any "Defeat" result, the Rebel Army suffers no losses. - §1.11.37

### Implementation notes

- `combat_result(modified_result)` reads the Defeat, Stalemate and Victory rows, and `combat_losses(result, modified_result, unit_count)` gives the losses for one force. `resolve_combat` calls `combat_losses` once for its fleets and once for its legions, as it already did inline.
- The revolt battle needs the same table for two armies, the Senate's and the rebel's, so this is split out first to keep that PR to the new rules.
- No behaviour change. Disaster and Standoff still come from the war and leader numbers before the table is read.

### Testing

674 backend tests pass and `mypy .` is clean across 456 files.


iamlogand (approved):
Thanks for improving this bit

PR #874, opened 2026-09-19
chore: rename opponent

### LINKED ISSUE #731: Replace generic uses of the word "opponent"
The word opponent should be reserved for the situation where one statesmen is personally opposed to another statesman. It's being used for influence peddling at the moment, so a different word should be used there instead.

iamlogand:
Closes #731 by renaming some things


ActualRory (approved):

ActualRory:
I know it's still a draft, but all looks good if you don't have more work to do here.

PR #876, opened 2026-09-20
chore: sentence case event names in event.json

iamlogand:
- Allied enthusiasm unchanged because it was already correct
- Not applicable to Verres because it's a proper noun


ActualRory (approved):
A bold change, as a general fan of capital letters it pains me to approve.
Alas.

